PR #2, opened 2019-09-19
First stab at an attempt to change the submodules.

porridge:


akirillov:
Merging it as CI issues are not related to submodules or repo layout. Thanks, @porridge

PR #3, opened 2019-09-19
[DCOS-58674] Automate creation of Google Spark Operator image

alembiewski:
This PR resolves [DCOS-58674: Automate creation of Google Spark Operator image](https://jira.mesosphere.com/browse/DCOS-58674)

Changes:
- Added `Dockerfile` for building Spark base image based on Spark 2.4.4 tarball;
- Added `Dockerfile` for building Google Spark Operator, using custom Spark base image;
- Added new `make` targets for building and pushing the Google Spark operator image to DockerHub repo

DockerHub repo needs to be created in scope of this task, so as soon is we agreed on repository's name, ticket will be created for IT.

- Tested locally with `make`
- As for now, base spark image tested only on `minikube` via `spark-submit` for sanity using the following command:
```
bin/spark-submit \                                                                                                              
   --master k8s://https://$(minikube ip):8443 \
   --deploy-mode cluster \
   --name spark-pi \
   --class org.apache.spark.examples.SparkPi \
   --conf spark.executor.instances=1 \
   --conf spark.kubernetes.container.image=lembiewski/spark-k8s:latest \
   --conf spark.kubernetes.driver.pod.name=spark-pi-driver \
 local:///opt/spark/examples/jars/spark-examples_2.11-2.4.4.jar
```


Makefile
```diff
@@ -0,0 +1,16 @@
+SPARK_ON_K8S_OPERATOR_REPO_NAME ?= mesosphere
+SPARK_ON_K8S_OPERATOR_IMAGE_NAME ?= spark-on-k8s-operator
+SPARK_ON_K8S_OPERATOR_FULL_IMAGE_NAME ?= $(SPARK_ON_K8S_OPERATOR_REPO_NAME)/$(SPARK_ON_K8S_OPERATOR_IMAGE_NAME)
+SPARK_ON_K8S_DOCKER_FILE_PATH ?= shared/spark-on-k8s-operator
+SPARK_DOCKER_IMAGE_NAME ?= "gcr.io/spark-operator/spark:v2.4.0"
```
akirillov: I'd suggest adding Apache Spark distribution tarball ourselves instead of using Google image

Makefile
```diff
@@ -0,0 +1,16 @@
...
+
+.PHONY: operator-docker-build
+operator-docker-build:
```
akirillov: how about renaming this and the target below to just `docker-build` and `docker-push`?

alembiewski: renamed as suggested

images/spark/Dockerfile
```diff
@@ -0,0 +1,52 @@
+ARG SPARK_HOME=/opt/spark
+
+FROM alpine:3.10.2 AS spark-dist
+
+ARG SPARK_DIST_URL="http://archive.apache.org/dist/spark/spark-2.4.4/spark-2.4.4-bin-hadoop2.7.tgz"
+ARG SPARK_HOME
```
akirillov: looks like a duplicate `ARG`

images/spark/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+COPY --from=spark-dist ${SPARK_HOME}/bin bin
+COPY --from=spark-dist ${SPARK_HOME}/sbin sbin
+COPY --from=spark-dist ${SPARK_HOME}/data data
+COPY --from=spark-dist ${SPARK_HOME}/examples examples
+COPY --from=spark-dist ${SPARK_HOME}/kubernetes/dockerfiles/spark/entrypoint.sh /opt/
+COPY --from=spark-dist ${SPARK_HOME}/kubernetes/tests tests
+
+# https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/591
+RUN rm jars/kubernetes-client-4.1.2.jar
+ADD https://repo1.maven.org/maven2/io/fabric8/kubernetes-client/4.4.2/kubernetes-client-4.4.2.jar jars
```
akirillov: 👍

Makefile
```diff
@@ -0,0 +1,16 @@
...
+SPARK_ON_K8S_DOCKER_FILE_PATH ?= images/operator/Dockerfile
+
+.PHONY: spark-build
+spark-build:
+	docker build -t ${SPARK_IMAGE_NAME} images/spark
+
+.PHONY: docker-build
+docker-build:
+	docker build -t ${SPARK_ON_K8S_OPERATOR_IMAGE_NAME} \
+	-f ${SPARK_ON_K8S_DOCKER_FILE_PATH} shared/spark-on-k8s-operator
```
akirillov: can you please rebase on the latest master because this path `shared/spark-on-k8s-operator` is just `spark-on-k8s-operator` now?

Makefile
```diff
@@ -0,0 +1,16 @@
+SPARK_IMAGE_NAME ?= mesosphere/spark-2.4.4-bin-hadoop2.7-k8s
+SPARK_ON_K8S_OPERATOR_IMAGE_NAME ?= mesosphere/spark-on-k8s-operator
+SPARK_ON_K8S_DOCKER_FILE_PATH ?= images/operator/Dockerfile
+
+.PHONY: spark-build
+spark-build:
+	docker build -t ${SPARK_IMAGE_NAME} images/spark
+
+.PHONY: docker-build
+docker-build:
```
akirillov: shouldn't this target depend on spark-build?

Makefile
```diff
@@ -0,0 +1,16 @@
+SPARK_IMAGE_NAME ?= mesosphere/spark-2.4.4-bin-hadoop2.7-k8s
+SPARK_ON_K8S_OPERATOR_IMAGE_NAME ?= mesosphere/spark-on-k8s-operator
+SPARK_ON_K8S_DOCKER_FILE_PATH ?= images/operator/Dockerfile
+
+.PHONY: spark-build
+spark-build:
+	docker build -t ${SPARK_IMAGE_NAME} images/spark
```
akirillov: do we want to `echo` the resulting image name to a file and use that file in other targets?

images/spark/Dockerfile
alembiewski: yes, it looks like it is, but this is actually the way to use global `ARG` in Docker: https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

alembiewski:
@akirillov, thanks for the review! Should we wait for CI checks to be green and then merge it?
Regarding your question: are you suggesting to have a single `Dockerfile` for Spark operator? 
> do we really need to depend on Spark image when we build Operator?

I guess so, we have dependency in `Dockerfile`: https://github.com/mesosphere/kudo-spark-operator/blob/7424ab24e6b72e6cbffaea06798827928b1d262e/images/operator/Dockerfile#L15

> It looks like the operator is pretty independent of Spark tarball.

Not quite sure I understand what you mean - operator is build from Spark image, which itself is built from tarball, therefore there are some dependencies. Maybe I'm missing something. Could you elaborate more on this one?

images/spark/Dockerfile
akirillov: got it, thanks. Although yet another question here: why do we need multi-stage build here? From what I can tell we just download and unpack a tarball, what are the benefits? Caching?

akirillov:
> Should we wait for CI checks to be green and then merge it?
For the time being [Konvoy KUDO 0.5.0 build](https://teamcity.mesosphere.io/viewLog.html?buildId=2309337&buildTypeId=Frameworks_DataServices_Kudo_Spark_Pr_SparkPrKonvoyKudoV050&guest=1) is sufficient

Regarding Spark Operator dependency:
- the Docker image depends on Spark image - true, but the operator code doesn't. Spark Operator is a standalone binary which doesn't invoke any of Spark native commands so there's no need in Spark distribution being available in the image
- it can be the case when by default Spark Operator uses the same image for both itself and for Spark Drivers

So let's keep it as-is for now and optimize later if needed.

images/spark/Dockerfile
alembiewski: Yes, in that particular case multi-stage build doesn't introduce any significant benefits, but from my perspective it has some pros:
- the final image is more lightweight, due to exclusion of the tools we need to download and verify Spark dist;
- we don't need to do manual cleanup - just pick what we need we need and discard all the rest.
But this approach has one big drawback - it leaves intermediate image, which needs to be cleaned manually. Therefore clean-up is needed one way or another. Given that, I'm considering to refactor the `Dockerfile` and perform build as single stage.

PR #5, opened 2019-09-26
[DCOS-58676] Build automation tooling

alembiewski:
Resolves [DCOS-58676: Build automation tooling](https://jira.mesosphere.com/browse/DCOS-58676)

Changes:
- Add Dockerfile for builder image with required tools;
- Add/extend `make` targets;
- Add tag to `spark-on-k8s-operator` image;
- Refactoring

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,13 +14,24 @@ NAMESPACE ?= spark-operator
 
 CLUSTER_TYPE ?= konvoy
 
-SPARK_IMAGE_NAME ?= mesosphere/spark-2.4.4-bin-hadoop2.7-k8s
+DOCKER_REPO_NAME ?= mesosphere
+
+SPARK_IMAGE_NAME ?= spark-2.4.4-bin-hadoop2.7-k8s
 SPARK_IMAGE_DOCKERFILE_PATH ?= $(ROOT_DIR)/images/spark
-SPARK_IMAGE_TAG ?= $(shell cat $(SPARK_IMAGE_DOCKERFILE_PATH)/Dockerfile | sha1sum  | cut -d ' ' -f1)
-SPARK_IMAGE_NAME_WITH_TAG ?= $(SPARK_IMAGE_NAME):$(SPARK_IMAGE_TAG)
-OPERATOR_IMAGE_NAME ?= mesosphere/kudo-spark-operator
-OPERATOR_VERSION ?= latest
-SPARK_ON_K8S_OPERATOR_DOCKERFILE_PATH ?= $(ROOT_DIR)/images/operator/Dockerfile
+SPARK_IMAGE_TAG ?= $(call get_sha1sum,$(SPARK_IMAGE_DOCKERFILE_PATH)/Dockerfile)
+SPARK_IMAGE_FULL_NAME ?= $(DOCKER_REPO_NAME)/$(SPARK_IMAGE_NAME):$(SPARK_IMAGE_TAG)
+
+OPERATOR_IMAGE_NAME ?= kudo-spark-operator
+OPERATOR_DOCKERFILE_PATH ?= $(ROOT_DIR)/images/operator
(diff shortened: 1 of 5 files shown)


images/builder/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM golang:1.13.0@sha256:de697ce5ae02f3d9a57b0603fbb648efadfa212727e702ad3a807b43eba7f6d6
+
+RUN curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.16.0/bin/linux/amd64/kubectl && \
```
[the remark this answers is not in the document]
alembiewski: Actually, `kubectl` is mandatory as well as `go` environment setup

Makefile
```diff
@@ -37,28 +49,60 @@ cluster-destroy:
...
+		-v $(ROOT_DIR)/admin.conf:/root/.kube/config \
+		-e SPARK_IMAGE=$(SPARK_IMAGE_FULL_NAME) \
+		-e SPARK_OPERATOR_IMAGE=$(SPARK_ON_K8S_OPERATOR_IMAGE_FULL_NAME) \
+		$(shell cat docker-builder) \
+		/bin/bash -c \
+		"kubectl config use-context kubernetes-admin@kudo-spark-operator && \
+		kubectl cluster-info && \
+		echo \$$SPARK_IMAGE && \
+		echo \$$SPARK_OPERATOR_IMAGE && \
+		/run-tests.sh"
```
akirillov: will that work well with CI in terms of all the environment variables passed to  `/run-tests.sh`?  We should either invoke `make test` from `/run-tests.sh` and do all the configuration here, or make sure CI will set all the same parameters needed for the script. The contract is that CI triggers only `/run-tests.sh`

Makefile
```diff
@@ -37,28 +49,60 @@ cluster-destroy:
...
+
+.PHONY: docker-builder
+docker-builder:
+	if [[ -z "$(shell docker images -q $(DOCKER_BUILDER_IMAGE_FULL_NAME))" ]]; then
+		docker build \
+			-t $(DOCKER_BUILDER_IMAGE_FULL_NAME) \
+			-f ${DOCKER_BUILDER_DOCKERFILE_PATH}/Dockerfile \
+			${DOCKER_BUILDER_DOCKERFILE_PATH}
+	fi
+	echo $(DOCKER_BUILDER_IMAGE_FULL_NAME) > $@
```
akirillov: it looks like this line can be moved inside the conditional above

Makefile
```diff
@@ -37,28 +49,60 @@ cluster-destroy:
...
+	echo "${SPARK_IMAGE_FULL_NAME}" > $@
 
 .PHONY: docker-build
 docker-build: spark-build
 docker-build:
-	docker build \
-		--build-arg SPARK_IMAGE=$(shell cat spark-build) \
-		-t ${SPARK_ON_K8S_OPERATOR_IMAGE_NAME} \
-		-f ${SPARK_ON_K8S_OPERATOR_DOCKERFILE_PATH} ${SPARK_OPERATOR_DIR}
+	if [[ -z "$(shell docker images -q $(SPARK_ON_K8S_OPERATOR_IMAGE_FULL_NAME))" ]]; then
```
akirillov: it's a neat move but don't we want to rely on `make` here and just `echo` the image name to the target file? Also, it doesn't look like Docker-related targets should be `PHONY`, we can introduce `clean-docker` target for the rebuild purposes.

alembiewski:
Thanks for the review @akirillov! All the comments have been addressed in accordance with CI contract as well as local development workfolw, thanks for the input!

alembiewski:
Thanks @akirillov ! It looks like this one could be merged to master. @rpalaznik, could you rebase on master and use docker builder to run tests in scope of #6?

PR #11, opened 2019-10-01
DCOS-59489 [DS] [Spark Operator] Update KUDO Spark Operator to support latest apiVersion

VladimirKravtsov:
**What is addressed in this PR?**
* synced mesosphere/spark-on-k8s-operator with the upstream
* updated spark-on-k8s-operator submodule in mesosphere/kudo-spark-operator to point to v1beta2-1.0.0-2.4.4 tag
* updated tooling to be able to build a working operator image which supports v1beta2 API version

Images build logs - [make_operator-build.log](https://github.com/mesosphere/kudo-spark-operator/files/3678844/make_operator-build.log)

kudo-spark-operator installation logs - [install_operator.log](https://github.com/mesosphere/kudo-spark-operator/files/3681171/install_operator.log)

The operator/controller logs - [logs-from-sparkoperator-in-spark-operator-sparkoperator-6bc599f75c-5xjff.txt](https://github.com/mesosphere/kudo-spark-operator/files/3681186/logs-from-sparkoperator-in-spark-operator-sparkoperator-6bc599f75c-5xjff.txt)

image

**How to verify?**

**_Prerequisite_**
* Konvoy cluster is running and up. 
* `kubectl` configured to talk to the Konvoy cluster
* Docker installed on your host.

Steps:
1) Build operator (using customized docker repository name) - 
 ```bash
make operator-build DOCKER_REPO_NAME=<your_own_public_repo_on_docker_hub>
```
2) Login to Docker Hub:
```bash
docker login
``` 
3) Push these just built images to Docker Hub -
```bash
make docker-push DOCKER_REPO_NAME=<your_own_public_repo_on_docker_hub>
```
4) Install the operator onto Konvoy cluster -
```bash
make install
```
5) Install a Spark application example 
```bash 
kubectl create -n spark-operator -f spec/spark-application.yaml
```
6) 
* Go to Konvoy/Kubernetes dashboard. 
* Pick `spark-operator` namespace.
* And verify pods states on `Pods` tab, check their logs. 

There is all should be green.

Cheers!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,14 +17,14 @@ KUBECONFIG ?= $(ROOT_DIR)/admin.conf
 
 DOCKER_REPO_NAME ?= mesosphere
 
-SPARK_IMAGE_NAME ?= spark-k8s
+SPARK_IMAGE_NAME ?= spark-2.4.4-bin-hadoop2.7-k8s
 SPARK_IMAGE_DOCKERFILE_PATH ?= $(ROOT_DIR)/images/spark
-SPARK_IMAGE_TAG ?= 2.4.4-hadoop2.7
+SPARK_IMAGE_TAG ?= $(call get_sha1sum,$(SPARK_IMAGE_DOCKERFILE_PATH)/Dockerfile)
 SPARK_IMAGE_FULL_NAME ?= $(DOCKER_REPO_NAME)/$(SPARK_IMAGE_NAME):$(SPARK_IMAGE_TAG)
 
 OPERATOR_IMAGE_NAME ?= kudo-spark-operator
 OPERATOR_DOCKERFILE_PATH ?= $(ROOT_DIR)/images/operator
-OPERATOR_VERSION ?= $(shell cat VERSION)
+OPERATOR_VERSION ?= $(call get_sha1sum,$(OPERATOR_DOCKERFILE_PATH)/Dockerfile)
 OPERATOR_IMAGE_FULL_NAME ?= $(DOCKER_REPO_NAME)/$(OPERATOR_IMAGE_NAME):$(OPERATOR_VERSION)
 
 DOCKER_BUILDER_IMAGE_NAME ?= spark-operator-docker-builder
@@ -108,7 +108,3 @@ clean-all:
 clean-docker:
 	rm -f *-build docker-builder
 
-.PHONY: version
-version:
-	echo spark-k8s version - "${SPARK_IMAGE_TAG}"
-	echo kudo-spark-operator version - "${OPERATOR_VERSION}"
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ docker-push:
 
 .PHONY: install
 install:
-	$(SCRIPTS_DIR)/install_operator.sh
+	OPERATOR_IMAGE_NAME=$(DOCKER_REPO_NAME)/$(OPERATOR_IMAGE_NAME) OPERATOR_VERSION=$(OPERATOR_VERSION) $(SCRIPTS_DIR)/install_operator.sh
 
 docker-builder:
 	docker build \
--- a/Makefile
+++ b/Makefile
@@ -17,14 +17,14 @@ KUBECONFIG ?= $(ROOT_DIR)/admin.conf
 
 DOCKER_REPO_NAME ?= mesosphere
 
-SPARK_IMAGE_NAME ?= spark-2.4.4-bin-hadoop2.7-k8s
+SPARK_IMAGE_NAME ?= spark-k8s
 SPARK_IMAGE_DOCKERFILE_PATH ?= $(ROOT_DIR)/images/spark
-SPARK_IMAGE_TAG ?= $(call get_sha1sum,$(SPARK_IMAGE_DOCKERFILE_PATH)/Dockerfile)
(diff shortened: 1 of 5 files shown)


VladimirKravtsov:
Hi Anton (@akirillov),

> How the changes were tested? 

I have updated description, where added `How to verify?` section.

> Have you tried deploying the new version and submitting Spark Application with `apiVersion: v1beta2`? 

Sure, I checked it on our `spec/spark-application.yaml` example -  it works. 

> Can you please attach controller logs as well?

I'm not sure it's helpful a lot, but I have added it to description.

Makefile
```diff
@@ -17,14 +17,14 @@ KUBECONFIG ?= $(ROOT_DIR)/admin.conf
...
+SPARK_IMAGE_NAME ?= spark-k8s
 SPARK_IMAGE_DOCKERFILE_PATH ?= $(ROOT_DIR)/images/spark
-SPARK_IMAGE_TAG ?= $(call get_sha1sum,$(SPARK_IMAGE_DOCKERFILE_PATH)/Dockerfile)
+SPARK_IMAGE_TAG ?= 2.4.4-hadoop2.7
 SPARK_IMAGE_FULL_NAME ?= $(DOCKER_REPO_NAME)/$(SPARK_IMAGE_NAME):$(SPARK_IMAGE_TAG)
 
 OPERATOR_IMAGE_NAME ?= kudo-spark-operator
 OPERATOR_DOCKERFILE_PATH ?= $(ROOT_DIR)/images/operator
-OPERATOR_VERSION ?= $(call get_sha1sum,$(OPERATOR_DOCKERFILE_PATH)/Dockerfile)
+OPERATOR_VERSION ?= $(shell cat VERSION)
```
akirillov: as discussed during the meeting if we want to fix the issue related to the fact that single Dockerfile checksum doesn't represent the full contents of the image: other files which copied to the image and operator git commit SHA. So to make this work properly we can stick to the following approach:
- use checksum of all files in the folder (for Operator image it should include parent Spark image checksum)
-  concatenate that computed checksum with Git commit SHA of Spark Operator submodule

Here's a possible implementation of the folder SHA:
```
# to avoid inconsistencies with files ordering in Linux/Unix systems the final checksum is
# calculated based on sorted checksums of independent files stored in a temporary file
files_checksum() {
    files=("$@")
    echo > tmp.checksum
    if [[ -n "$@" ]]; then
        for file in "${files[@]}"; do
            sha1sum ${file} | cut -d' ' -f1 >> tmp.checksum
        done
    fi
    sort tmp.checksum | sha1sum | cut -d' ' -f1
}
```

which can be invoked as follows: `$(files_checksum $(find "${OPERATOR_DOCKERFILE_PATH}/shared" -type f))`

VladimirKravtsov: Anton (@akirillov), thanks for spent additional time to discuss this (minor) stuff during the meeting! PTAL^

scripts/install_operator.sh
```diff
@@ -6,9 +6,15 @@ SPECS_DIR="$(dirname ${SCRIPT_DIR})/specs"
 
 NAMESPACE=${NAMESPACE:-spark-operator}
 
-OPERATOR_NAME=${OPERATOR_NAME:-spark-operator}
+if [ -f "${SCRIPT_DIR}"/../operator-build ]; then
+  echo "operator-build file is found. Parse OPERATOR_IMAGE_NAME and OPERATOR_VERSION from there."
+  OPERATOR_IMAGE_NAME=$(awk -F':' '{printf $1}' <"${SCRIPT_DIR}"/../operator-build)
+  OPERATOR_VERSION=$(awk -F':' '{printf $2}' <"${SCRIPT_DIR}"/../operator-build)
```
akirillov: in case a user executes the following command, provided variables will be ignored in case file `operator-build` is present: 
```
OPERATOR_IMAGE_NAME=user/spark-operator OPERATOR_VERSION=latest ./scripts/install_operator.sh
```
can we rely on the values provided in Makefile so the script doesn't make assumptions about any files created by external tools and directories layout of the project?

VladimirKravtsov: it's good point about override user defined env vars, thanks for catch it. Fixed.

> can we rely on the values provided in Makefile so the script doesn't make assumptions about any files created by external tools and directories layout of the project?

make file generates tag versions in runtime, no chance to get it without run some `make` target.

akirillov: but that's the whole point, isn't it? So here are the scenarios:
- a user runs `make install` and values are propagated
- a user runs `install_operator.sh` and defaults are applied
- a user runs `OPERATOR_IMAGE_NAME=user/spark-operator OPERATOR_VERSION=latest install_operator.sh` and gets her values propagated
- (OR) `make install OPERATOR_IMAGE_NAME=user/spark-operator OPERATOR_VERSION=latest`

so with the current approach what doesn't look great is that:
- the script is aware of some file and its location (rename the file, and the script stops working)
- the script is aware of the structure of this file and knows how to parse it (same problem as above)

So I'd suggest removing this newly added logic and rely on `make` or a user for passing necessary parameters at the moment of execution. If a user wants to rerun the script, she can run `make install` one more time. 

What we can do to improve it - is to remove the default values in the script and add a check for those variables presence. If variables are present - we add Helm/KUDO image parameters, if not - rely on defaults specified in Chart/Operator. 
WDYT?

PR #14, opened 2019-10-03
[DCOS-59174] [DS] [Spark Operator] Possibility to integrate Spark History Server with Spark via PVC (POC)

farhan5900:
Resolves [DCOS-59174: Integrate Spark History Server](https://jira.mesosphere.com/browse/DCOS-59174) and [DCOS-60038: S3 as storage backend](https://jira.mesosphere.com/browse/DCOS-60038)

Introduces Deployment and Service for the Spark History Server as a Kudo operator templates. There is an option to use PVC (It should support ReadWriteMany mode) for the storage of History Server events log. It also enable a support to use S3 bucket as storage for events log.

Introducing Spark History Server to the Spark Kudo Operator

Here are the steps to run Spark Operator with Spark History Server enabled backed by S3 for logs.

* Build and push an image:
`make docker-push DOCKER_REPO_NAME=xyz`

* Set the value for following parameters of `kudo-operator/params.yaml`.

```
enableHistoryServer: true
historyServerFsLogDirectory: "s3a://<BUCKET_NAME>/<FOLDER>"
historyServerOpts: "-Dspark.hadoop.fs.s3a.access.key=<AWS_ACCESS_KEY_ID> -Dspark.hadoop.fs.s3a.secret.key=<AWS_SECRET_ACCESS_KEY> -Dspark.hadoop.fs.s3a.impl=org.apache.hadoop.fs.s3a.S3AFileSystem"
```
Replace all the angle bracketed varaibles with actual values. If you are using `kubectl kudo` directly you can pass these as parameters while installing Spark Operator.

* Install the Spark Operator:
`make install DOCKER_REPO_NAME=xyz OPERATOR_IMAGE_NAME=kudo-spark-operator OPERATOR_VERSION=123` NAMESPACE=spark-operator

* In the `specs/spark-application.yaml` add these configurations under `spec.sparkConf`:
```
"spark.eventLog.enabled": "true"
"spark.eventLog.dir": "s3a://<BUCKET_NAME>/<FOLDER>"
"spark.hadoop.fs.s3a.access.key": "<AWS_ACCESS_KEY_ID>"
"spark.hadoop.fs.s3a.secret.key": "<AWS_SECRET_ACCESS_KEY>"
```
Also modify the `spec.image` with the latest build image of spark.

* Run the SparkApplication:
`kubectl apply --namespace=spark-operator -f specs/spark-application.yaml`

* To expose the SHS UI you can run following command, replace the actual pod name:
`kubectl port-forward <spark-history-server-pod-name> 18080:18080`
You can access the UI at this: `http://localhost:18080`.


specs/spark-historyserver.yaml
```diff
@@ -0,0 +1,42 @@
...
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: historyserver
+        app.kubernetes.io/version: v2.4.4-v1beta2
+    spec:
+      serviceAccountName: spark-operator-sparkoperator
+      containers:
+       - name: historyserver
+         image: farhan5900/kudo-spark-operator:327b0d0bbc4444f01c565c02618dd90c5f2d4a03
```
alembiewski:
```suggestion
         image: mesosphere/kudo-spark-operator:latest
```

alembiewski: let's use the image from `mesosphere`'s repo since it's already published

scripts/install_operator.sh
```diff
@@ -23,3 +23,6 @@ else
 fi
 
 kubectl apply --namespace "${NAMESPACE}" -f ${SPECS_DIR}/spark-driver-rbac.yaml
+kubeclt apply --namespace "${NAMESPACE}" -f ${SPECS_DIR}/spark-events-pvc.yaml
+kubeclt apply --namespace "${NAMESPACE}" -f ${SPECS_DIR}/spark-history-server.yaml
+kubeclt apply --namespace "${NAMESPACE}" -f ${SPECS_DIR}/spark-historyserver-service.yaml
```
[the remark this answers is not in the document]
alembiewski: +1 here, but I would propose to use `kebab-case` and to separate words with dashes (i.e. `history-server`)

kudo-operator/params.yaml
```diff
@@ -92,6 +92,21 @@ ingressUrlFormat:
...
+  description: "Start Spark History Server"
+  default: true
+  displayName: "Enable Spark History Server"
+
+historyServerFsLogDirectory:
+  description: "Spark EventLog Directory from which to load events for prior Spark job runs (e.g., hdfs://hdfs/ or s3a://path/to/bucket). Note: You'd typically want to set this to point to distributed/HA storage"
+  default: "/opt/spark/work-dir"
+  displayName: "Spark History Server Event Log Directory"
+
+storageClassName:
```
akirillov: it makes sense to add something like `createPersistentVolumeClaim` which defaults to `true` iff `historyServerFsLogDirectory` is not S3 or HDFS - can we support this kind of logic in `spark-events-pvc.yaml`?

kudo-operator/templates/spark-events-pvc.yaml
```diff
@@ -0,0 +1,14 @@
+{{ if .Params.enableHistoryServer }}
```
akirillov: based on the previous comment, let's add `createPersistentVolumeClaim` to the condition

kudo-operator/templates/spark-events-pvc.yaml
```diff
@@ -0,0 +1,14 @@
...
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: spark-events-pvc
+  namespace: {{ .Namespace }}
+spec:
+  storageClassName: {{ .Params.storageClassName }}
```
akirillov: just wondering if we can omit always explicit configuration and rely on [DefaultStorageClass](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#defaultstorageclass) admission controller?

kudo-operator/params.yaml
```diff
@@ -92,6 +92,21 @@ ingressUrlFormat:
...
+  description: "Storage class for provisioning a persistent volume used for Spark Event Logging."
+  default: "awsebscsiprovisioner"
+  displayName: "Storage Class Name"
+
```
akirillov: let's add `persistentVolumeClaimName` to allow users provide their own PVCs when `createPersistentVolumeClaim` is `false`.  When `createPersistentVolumeClaim` is `true` then `persistentVolumeClaimName` should be used when creating a PVC. Also, users might create an S3 PVC so it makes sense to standardize `eventLog dir` mounting and configuration via PVCs.

kudo-operator/params.yaml
```diff
@@ -92,6 +92,21 @@ ingressUrlFormat:
...
+  displayName: "Enable Spark History Server"
+
+historyServerFsLogDirectory:
+  description: "Spark EventLog Directory from which to load events for prior Spark job runs (e.g., hdfs://hdfs/ or s3a://path/to/bucket). Note: You'd typically want to set this to point to distributed/HA storage"
+  default: "/opt/spark/work-dir"
+  displayName: "Spark History Server Event Log Directory"
+
+storageClassName:
+  description: "Storage class for provisioning a persistent volume used for Spark Event Logging."
+  default: "awsebscsiprovisioner"
```
akirillov: can we either rely on `DefaultStorageClass` admission controller or request this parameter from a user if `DefaultStorageClass` is not configured and `createPersistentVolumeClaim` is `true`? It would be really great to avoid binding to a single cloud provider so probably we can just make PVC creation external to KUDO Operator if there's no way to avoid explicit configuration?

kudo-operator/templates/spark-events-pvc.yaml
zmalik: I think it makes sense to use by default the `defaultStorageClass` but leave the door open for any other storage class using parameters 
https://github.com/kudobuilder/operators/blob/master/repository/kafka/operator/templates/statefulset.yaml#L157

kudo-operator/params.yaml
```diff
@@ -92,6 +92,21 @@ ingressUrlFormat:
...
+  description: "Start Spark History Server"
+  default: true
+  displayName: "Enable Spark History Server"
+
+historyServerFsLogDirectory:
+  description: "Spark EventLog Directory from which to load events for prior Spark job runs (e.g., hdfs://hdfs/ or s3a://path/to/bucket). Note: You'd typically want to set this to point to distributed/HA storage"
+  default: "/opt/spark/work-dir"
+  displayName: "Spark History Server Event Log Directory"
+
+persistentVolumeClaimName:
```
akirillov: I'd suggest making the name more explicit, something like `historyServerPVC[Name]` to make it clear that PVC is related to SHS.

kudo-operator/params.yaml
```diff
@@ -92,6 +92,21 @@ ingressUrlFormat:
...
+  description: "External Persistent Volume Claim Name"
```
akirillov: Let's extend the description so that it includes the purpose. Something like this: `External Persistent Volume Claim Name used for Spark event logs storage and History Server`

kudo-operator/params.yaml
```diff
@@ -92,6 +92,21 @@ ingressUrlFormat:
...
+  default: ""
+  displayName: "Persistent Volume Claim Name"
```
akirillov:
```suggestion
  displayName: "Spark History Server Persistent Volume Claim Name"
```

kudo-operator/templates/spark-history-server-deployment.yaml
```diff
@@ -0,0 +1,53 @@
+{{ if .Params.enableHistoryServer }}
```
akirillov: Let's use hyphens when we template conditionals, we use it throughout the rest of the templates and IMO it improves readability and allows distinguishing control structures from variables:
```
{{- if .Params.enableHistoryServer }}
<spec>
{{- end }}
```

kudo-operator/templates/spark-history-server-service.yaml
```diff
@@ -0,0 +1,18 @@
+{{ if .Params.enableHistoryServer }}
```
akirillov:
```suggestion
{{- if .Params.enableHistoryServer }}
```

kudo-operator/templates/spark-history-server-service.yaml
```diff
@@ -0,0 +1,18 @@
...
+    app.kubernetes.io/instance: {{ .Name }}
+spec:
+  ports:
+  - port: 18080
+    name: history-server
+  selector:
+    app.kubernetes.io/name: history-server
+    app.kubernetes.io/instance: {{ .Name }}
+    app.kubernetes.io/version: {{ .Params.operatorVersion }}
+{{ end }}
```
akirillov:
```suggestion
{{- end }}
```

kudo-operator/templates/spark-events-pvc.yaml
akirillov: thanks for the pointer, @zmalik. With Spark, it turns out that `StorageClass` is too low-level abstraction because users require to mount volumes to every Spark Application they submit. So our thinking here is to provide instructions on how to create PVC and specify its name here instead of storage class and in volume mounts of each application.

images/spark/Dockerfile
```diff
@@ -34,6 +34,10 @@ WORKDIR ${SPARK_HOME}
 RUN rm jars/kubernetes-client-4.1.2.jar
 ADD https://repo1.maven.org/maven2/io/fabric8/kubernetes-client/4.4.2/kubernetes-client-4.4.2.jar jars
 
+# Jars for S3 access support in case of Spark History Server
+ADD http://central.maven.org/maven2/org/apache/hadoop/hadoop-aws/2.7.3/hadoop-aws-2.7.3.jar jars
+ADD http://central.maven.org/maven2/com/amazonaws/aws-java-sdk/1.7.4/aws-java-sdk-1.7.4.jar jars
```
akirillov: IMO it's better to include them in our Spark distribution tarball, but it'snot a showstopper, we can address it later.

kudo-operator/params.yaml
```diff
@@ -92,6 +92,21 @@ ingressUrlFormat:
...
   displayName: "Ingress URL format"
 
+enableHistoryServer:
+  description: "Start Spark History Server"
+  default: true
+  displayName: "Enable Spark History Server"
+
+historyServerFsLogDirectory:
+  description: "Spark EventLog Directory from which to load events for prior Spark job runs (e.g., hdfs://hdfs/ or s3a://path/to/bucket). Note: You'd typically want to set this to point to distributed/HA storage"
+  default: "/opt/spark/work-dir"
```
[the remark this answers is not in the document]
akirillov: I think this comment is still valid

kudo-operator/templates/spark-history-server-deployment.yaml
```diff
@@ -0,0 +1,64 @@
...
+        imagePullPolicy: {{ .Params.imagePullPolicy }}
+        command: ["/opt/spark/sbin/start-history-server.sh"]
+        env:
+        - name: SPARK_NO_DAEMONIZE
+          value: "true"
+        - name: AWS_ACCESS_KEY_ID
+          valueFrom:
+            secretKeyRef:
+              name: s3-secret
+              key: accessKeyID
```
akirillov: I'm afraid, this is not going to work in case the secret is not present and Spark History is enabled

kudo-operator/templates/spark-history-server-deployment.yaml
```diff
@@ -0,0 +1,64 @@
...
+        - name: AWS_SECRET_ACCESS_KEY
+          valueFrom:
+            secretKeyRef:
+              name: s3-secret
+              key: secretAccessKey
+        - name: SPARK_HISTORY_OPTS
```
akirillov: how about adding only this single field to `params.yaml` and let users specify all the options they want? So it can look like that:
```
- name: SPARK_HISTORY_OPTS
  value: "-Dspark.history.fs.logDirectory={{ .Params.historyServerFsLogDirectory }} {{.Params.historyServerOptions }}"
```

akirillov: using secrets is a nice approach though, but I don't think it must be done in the scope of this PR. Most of the AWS users rely on IAM and instance profiles for access management so it doesn't seem like the best option for them. And S3 is not the only supported storage... So let's skip the secrets for now and allow users to provide them via `{{.Params.historyServerOptions }}`

kudo-operator/templates/spark-history-server-deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+      containers:
+      - name: history-server
+        image: {{ .Params.operatorImageName }}:{{ .Params.operatorVersion }}
+        imagePullPolicy: {{ .Params.imagePullPolicy }}
+        command: ["/opt/spark/sbin/start-history-server.sh"]
+        env:
+        - name: SPARK_NO_DAEMONIZE
+          value: "true"
+        - name: SPARK_HISTORY_OPTS
+          value: "{{ .Params.historyServerOpts }} -Dspark.history.fs.logDirectory={{ .Params.historyServerFsLogDirectory }} -Dspark.history.fs.cleaner.enabled={{ .Params.historyServerCleanerEnabled }} -Dspark.history.fs.cleaner.interval={{ .Params.historyServerCleanerInterval }} -Dspark.history.fs.cleaner.maxAge={{ .Params.historyServerCleanerMaxAge }}"
```
akirillov: I think it's better to move `{{ .Params.historyServerOpts }}` to the end, so it acts as an override for the users too but it's up to you

farhan5900: I thought we should give preference to explicitly provided options.

PR #15, opened 2019-10-04
DCOS-59602 [Spark Operator] Add CRDss templates to kudo-spark-operator definition

VladimirKravtsov:
1) Resolves [DCOS-59602: Add CRDs templates to kudo-spark-operator definition](https://jira.mesosphere.com/browse/DCOS-59602)
1) Upgrade to `KUDO 0.7.3` (it's vital to have ability install CRDs within the operator). If you would like better understand the topic, please, take a look on the [message](https://mesosphere.slack.com/archives/CKKDWQDGW/p1570457285154700) and the [thread](https://mesosphere.slack.com/archives/CKKDWQDGW/p1570457499156000) in Slack.
1) Fixes `remove_operator.py`: 
   - Removed deletion of `planexecutions.kudo.dev`, because that Kind no longer exists with KUDO 0.7.3;
   - Fix ignoring specified `NAMESPACE` during resources deletion.

Having `kudo-spark-operator` installed, these changes introduce ability to deploy Spark jobs using `SparkApplication` and `ScheduledSparkApplication` kinds, defined in CRDs.

#### Prerequisite
- Konvoy cluster up and running.
- kubectl.
- KUDO-cli **v0.7.3**.
- Python3.
**_if you would like use cluster with any previous version of KUDO installed, then you need delete KUDO version beforehand._**
#### How to remove KUDO from cluster? 
1) `kubectl kudo init --dry-run -o yaml | kubectl delete -f -`
2) Delete all KUDO's CRDs - `kubectl delete crd *.kudo.dev`
3) `kubectl kudo init -wait`

#### How to verify the PR changes
1) Install/upgrade `kudo-cli` to [**0.7.3 version**](https://github.com/kudobuilder/kudo/releases/tag/v0.7.3). **_Critically important! [here](https://mesosphere.slack.com/archives/CKKDWQDGW/p1570457285154700) is why_**
1) Install `kudo-spark-operator` with via `NAMESPACE=<desired_namespace>  ./scripts/install_operator.sh`
1) Check that CRDs are installed - `kubectl get crd | grep spark`. Output should be like this - 
```bash
$ kubectl get crd | grep spark
scheduledsparkapplications.sparkoperator.k8s.io   2019-10-08T14:03:52Z
sparkapplications.sparkoperator.k8s.io            2019-10-08T14:03:53Z
```
1) Look into Kubernetes Dashboard, waiting Spark driver is up and running (green).
1) Deploy `SparkApplication` example -
`kubectl -n <desired_namepace> apply -f ./specs/spark-application.yaml`
1) Look into Kubernetes Dashboard, waiting Spark Job is up and running (green).
1) Delete operator `NAMESPACE=<>desired_namespace> ./scripts/remove_operator.py`
1) Check that CRDs are removed from cluster now - `kubectl get crd | grep spark`


alembiewski:
LGTM!

kudo-operator/operator.yaml
```diff
@@ -9,6 +9,7 @@ url: https://spark.apache.org/
 tasks:
   deploy:
     resources:
+      - crds.yaml
```
akirillov: looks like `crd-cleanup-job.yaml` needs to be added as well

scripts/remove_operator.py
```diff
@@ -32,3 +32,5 @@ def delete_resource(api, resource):
     delete_resource("operatorversion.kudo.dev", version)
 
 delete_resource("operator.kudo.dev", "spark")
+delete_resource("crd", "sparkapplications.sparkoperator.k8s.io")
```
samvantran: Not in scope of this PR but since we're using python, maybe we want to add `retrying` for these deletes in case of network issues

kudo-operator/templates/crds.yaml
```diff
@@ -0,0 +1,5091 @@
...
+                                      - type: integer
+                                  scheme:
+                                    type: string
+                                required:
+                                  - port
+                                type: object
+                              tcpSocket:
+                                properties:
+                                  host:
+
```
samvantran: should we break these two crds up into their own files so its easier to know which file to edit if we ever need to? Or is it if we ever change one value for `sparkapplications` we'll likely change it for `scheduledsparkapplications` too?

kudo-operator/operator.yaml
VladimirKravtsov: the `crd-cleanup-job.yaml` file contains Kubernetes job for CRDs removing and it definitely shouldn't be added in `deploy` plan. But in our case the CRDs, and other resources are listed here, are removing within operator - so, the file `crd-cleanup-job.yaml` no need here at all. I removed it.

kudo-operator/templates/crds.yaml
VladimirKravtsov: These CRDs shouldn't be ever edited manually.  We are discussing with @akirillov approach how to maintain the yaml. In general, it should be auto-generated like described [here](https://github.com/mesosphere/spark-on-k8s-operator/blob/master/docs/developer-guide.md)

VladimirKravtsov:
Thank you, All, for yours preliminaries reviews! Once I became unblock, I did final changes. PTAL^

kudo-operator/templates/crds.yaml
samvantran: ah, autogenerated. okay thanks

PR #29, opened 2019-10-21
DCOS-60183: Teamcity test output fix

rpalaznik:
Resolves [DCOS-60183](https://jira.mesosphere.com/browse/DCOS-60183)
Applies `go-test-teamcity` to the `go test ...` output

Until Teamcity is updated to a version that supports `go test` out of the box (if ever), we need a workaround to achieve the following:
- Recognize and list individual tests in a build
- Build status should reflect test run results

Both locally and in Teamcity

### DIFF
--- a/tests/basic_test.go
+++ b/tests/basic_test.go
@@ -5,6 +5,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"testing"
+	"time"
 )
 
 func TestMain(m *testing.M) {
@@ -75,7 +76,7 @@ func TestJobSubmission(t *testing.T) {
 		t.Error(err.Error())
 	}
 
-	err = spark.WaitUntilSucceeded(job)
+	err = spark.WaitUntilSucceeded(10*time.Minute, job)
 	if err != nil {
 		t.Error(err.Error())
 	}
--- a/tests/basic_test.go
+++ b/tests/basic_test.go
@@ -5,7 +5,6 @@ import (
 	log "github.com/sirupsen/logrus"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"testing"
-	"time"
 )
 
 func TestMain(m *testing.M) {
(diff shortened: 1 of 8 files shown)


tests/basic_test.go
```diff
@@ -75,7 +76,7 @@ func TestJobSubmission(t *testing.T) {
 		t.Error(err.Error())
 	}
 
-	err = spark.WaitUntilSucceeded(job)
+	err = spark.WaitUntilSucceeded(10*time.Minute, job)
```
akirillov: should we declare `10*time.Minute` as a constant?

tests/run.sh
```diff
@@ -9,5 +9,17 @@ set -e
 # - SPARK_VERSION
 # Default values can be looked up in tests/utils/common.go
 pushd $TEST_DIR
-go test -v -count=1 -timeout 30m .
+
+if [ -n "$TEAMCITY_VERSION" ]; then
+  # Running in Teamcity, therefore we need to convert `go test` output accordingly
+  if command -v go-test-teamcity; then
+    go test -v -count=1 -timeout 30m . | go-test-teamcity
```
akirillov: `go test -v -count=1 -timeout 30m` repeated three times here, shall we extract it to a variable?

also, it would be great to make `timeout` configurable and allow overriding it via an environment variable.

tests/run.sh
```diff
@@ -9,5 +9,17 @@ set -e
...
+  else
+    go test -v -count=1 -timeout 30m . | docker run -i xjewer/go-test-teamcity
```
akirillov: it would be really great to avoid this dependency. as we're already installing the needed library in `docker-builder`, shall we use it instead?

zmalik: we should just aim to output results using junit xml format that the majority of CI systems already support it.
TeamCity has a build feature around XML processing which we can use of instead adding dependencies here
image

tests/basic_test.go
rpalaznik: Done

tests/run.sh
rpalaznik: I removed the usage of the 3rd party docker image. As for junit xml format, I don't think `go test` supports it out of the box, so it would be changing one dependency for another. 

Also I heard from Uladzimir, that there is a plan to update Teamcity soon, so this workaround could be dropped in future.

PR #35, opened 2019-10-23
DCOS-58677: Smoke tests

rpalaznik:
Resolves [DCOS-58677](https://jira.mesosphere.com/browse/DCOS-58677)

Adds two smoke tests:

**Shuffle app smoke test**
Verifies number of executors
Verifies driver log content

**Mock task runner**
Starts and kills the job
Verifies no pods left in the namespace

With `make cluster-create` and `make test`


tests/smoke_test.go
```diff
@@ -0,0 +1,131 @@
...
+	"time"
+)
+
+func TestShuffleApp(t *testing.T) {
+	spark := utils.SparkOperatorInstallation{}
+	err := spark.InstallSparkOperator()
+	defer spark.CleanUp()
+
+	if err != nil {
+		t.Fatal(err)
```
akirillov: does this fail a single test or the whole run?

tests/smoke_test.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	err = spark.WaitForJobState(job, v1beta2.RunningState, 5*time.Minute)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Wait for correct number of executors to show up
+	err = utils.Retry(30*time.Second, time.Second, func() error {
```
akirillov: can we extract timeouts either to a constant or to some structure with timeouts per test?

tests/smoke_test.go
```diff
@@ -0,0 +1,131 @@
...
```
akirillov: same comment here regarding extracting timeouts to constants or config structure

tests/smoke_test.go
```diff
@@ -0,0 +1,131 @@
...
+			return errors.New(fmt.Sprintf("The number of executors is %d, but %d is expected", len(executors), expectedExecutorCount))
+		}
+		return nil
+	})
+	if err != nil {
+		t.Error(err)
+	}
+
+	// Terminate the job while it's running
+	spark.DeleteJob(job)
```
akirillov: as we delete the job and verify the cleanup, shall the test name reflect that? like `testRunningAppDeletion` or similar?

tests/templates/spark-shuffle-job.yaml
```diff
@@ -0,0 +1,38 @@
...
+  image: {{ .Image }}
+  imagePullPolicy: Always
+  mainClass: ShuffleApp
+  mainApplicationFile: "https://infinity-artifacts.s3.amazonaws.com/scale-tests/dcos-spark-scala-tests-assembly-2.4.0-20190325.jar"
+  arguments:
+    - "4"
+    - {{ index .Params "group_count" }}
+    - "100"
+    - "4"
+    - "1500"
```
akirillov: do we want to parameterize the rest of the arguments so we can vary them from the test code only?

tests/utils/k8s.go
```diff
@@ -66,9 +69,64 @@ func DropNamespace(clientSet *kubernetes.Clientset, name string) error {
 	})
 }
 
+func getPodLog(clientSet *kubernetes.Clientset, namespace string, pod string, lines int64) (string, error) {
```
akirillov: I think `lines` is a bit confusing name for a variable meant for `tail` operation, can we rename it to something like `tailLines`?

tests/utils/k8s.go
```diff
@@ -66,9 +69,64 @@ func DropNamespace(clientSet *kubernetes.Clientset, name string) error {
...
+	logBuffer := new(bytes.Buffer)
+	_, err = io.Copy(logBuffer, logSteam)
+	if err != nil {
+		return "", err
+	}
+
+	return logBuffer.String(), nil
+}
+
+func podLogContains(clientSet *kubernetes.Clientset, namespace string, pod string, text string) (bool, error) {
```
akirillov: is it an optimization to aviod getting full log from a pod? what would be the drawback of `return strings.Contains(getPodLog(<parameters>), text)`?

tests/utils/k8s.go
```diff
@@ -66,9 +69,64 @@ func DropNamespace(clientSet *kubernetes.Clientset, name string) error {
...
+	}
+
+	if err = scanner.Err(); err != nil {
+		return false, err
+	} else {
+		return false, nil
+	}
+}
+
+func logPodLogTail(clientSet *kubernetes.Clientset, namespace string, pod string, lines int64) error {
```
akirillov: and here, `lines` fits great as a name because the function name implies that this is the number of lines to tail

tests/utils/spark_operator.go
```diff
@@ -94,12 +106,21 @@ func (spark *SparkOperatorInstallation) CleanUp() {
...
+		panic(err.Error())
+	}
+
+	return operator.NewForConfig(config)
 }
 
 func (spark *SparkOperatorInstallation) waitForInstanceStatus(targetStatus string) error {
 	log.Infof("Waiting for %s/%s to reach status %s", spark.Namespace, spark.InstanceName, targetStatus)
-	return retry(3*time.Minute, 1*time.Second, func() error {
+	return Retry(3*time.Minute, 1*time.Second, func() error {
```
akirillov: should the timeout be passed as a parameter to `waitForInstanceStatus`?

tests/utils/spark_operator.go
```diff
@@ -115,6 +136,49 @@ func (spark *SparkOperatorInstallation) getInstanceStatus() (string, error) {
...
+	})
+
+	if err == nil {
+		log.Infof("Spark application %s is now %s", job.Name, state)
+	}
+
+	return err
+}
+
+func (spark *SparkOperatorInstallation) JobExecutors(job SparkJob) (map[string]v1beta2.ExecutorState, error) {
```
akirillov: function name looks confusing given that the returned entity is `ExecutorState`. Is the function a getter? Let's make this fact explicit if so

tests/templates/spark-mock-task-runner-job.yaml
```diff
@@ -0,0 +1,35 @@
...
+spec:
+  type: Scala
+  mode: cluster
+  image: {{ .Image }}
+  imagePullPolicy: Always
+  mainClass: MockTaskRunner
+  mainApplicationFile: "https://infinity-artifacts.s3.amazonaws.com/scale-tests/dcos-spark-scala-tests-assembly-2.4.0-20190325.jar"
+  arguments:
+    - "1"
+    - "600"
```
akirillov: does it make sense to parameterize the arguments?

tests/smoke_test.go
rpalaznik: Just this test

tests/smoke_test.go
rpalaznik: done

tests/smoke_test.go
rpalaznik: done

tests/smoke_test.go
rpalaznik: Makes sense, I've renamed the tests

tests/templates/spark-shuffle-job.yaml
rpalaznik: ok, i've parameterized all arguments for this app

tests/utils/k8s.go
rpalaznik: This is an optimization, as I have no idea how big the logs could be. RAM consumption is the only drawback I think.

tests/templates/spark-mock-task-runner-job.yaml
rpalaznik: I don't know. If this job will be reused in a new test with different arguments, it's fairly easy to refactor for parameterization in my opinion.

tests/utils/spark_operator.go
rpalaznik: renamed the method

rpalaznik:
@akirillov could you take another look?

> can we please organize retry timeouts somehow so we don't use magic numbers which differ from case to case? 

I've split retry function into two, with and without explicit timeout. The global timeout is 10 minutes now, which is fine for all current cases. In future tests new timeout constants could be added, if needed.

tests/utils/k8s.go
akirillov: thanks for addressing that!

tests/templates/spark-mock-task-runner-job.yaml
akirillov: makes sense, let's address that when it's needed

PR #45, opened 2019-10-30
DCOS-60350: Integration tests for service account creation

rpalaznik:
Resolves [DCOS-60350](https://jira.mesosphere.com/browse/DCOS-60350)
Adds tests for the following cases related to service account creation
- Verify that default configuration creates service accounts for the operator and spark jobs
- Verify that createOperatorServiceAccount=false configuration works with a provided service account
- Verify that createSparkServiceAccount=false configuration works with a provided service account

In a stand-alone AWS k8s cluster, with `make test`.

### DIFF
--- a/tests/security_test.go
+++ b/tests/security_test.go
@@ -1,0 +1,143 @@
+package tests
+
+import (
+	"github.com/mesosphere/kudo-spark-operator/tests/utils"
+	log "github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"testing"
+)
+
+func TestDefaultServiceAccounts(t *testing.T) {
+	const driverSA = "spark-service-account"
+	const operatorSA = "spark-operator-service-account"
+	spark := utils.SparkOperatorInstallation{
+		Namespace: "test-default-service-accounts",
+		Params: map[string]string{
+			"createOperatorServiceAccount": "true",
+			"createSparkServiceAccount":    "true",
+			"operatorServiceAccountName":   operatorSA,
+			"sparkServiceAccountName":      driverSA,
+		},
+	}
+
+	err := testServiceAccounts(spark,
+		utils.DefaultInstanceName+"-"+operatorSA,
+		utils.DefaultInstanceName+"-"+driverSA)
+	if err != nil {
+		t.Fatal(err.Error())
+	}
+}
+
+func TestSparkServiceAccount(t *testing.T) {
+	const namespace = "test-provided-spark-sa"
+	const driverSA = "spark-driver"
+	const operatorSA = "spark-operator-service-account"
+
+	// Prepare a namespace with a service account
+	err := createNamespaceAndServiceAccount(namespace, driverSA)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	spark := utils.SparkOperatorInstallation{
+		Namespace:            namespace,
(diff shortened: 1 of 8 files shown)


tests/security_test.go
```diff
@@ -0,0 +1,143 @@
+package tests
+
+import (
+	"github.com/mesosphere/kudo-spark-operator/tests/utils"
+	log "github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"testing"
+)
+
+func TestDefaultServiceAccounts(t *testing.T) {
```
VladimirKravtsov: Looks like these three tests methods(TestOperatorServiceAccount, TestSparkServiceAccount and this one) contains same logic. mb it’s make sense to extract a template method and invoke it with appropriate parameters in the tests methods, WDYT?

VladimirKravtsov: Also, perhaps, it makes sense consider the single test method instead of three separate, it could looks like
```suggestion
func TestServiceAccounts(t *testing.T) {
        t.Run("TestDefaultServiceAccounts", testServiceAccount(namespace1, driver1, operator1, t));
        t.Run("TestOperatorServiceAccount", testServiceAccount(namespace2, driver2, operator2, t));
        t.Run("TestSparkServiceAccount", testServiceAccount(namespace3, driver3, operator3, t));
}

func testServiceAccount(namespace string, driverSA string, operatorSA string, t *testing.T) {
     
    return func(t *testing.T) {
 // the test method template
 }
}
```

tests/smoke_test.go
```diff
@@ -55,10 +55,13 @@ func TestShuffleAppDriverOutput(t *testing.T) {
 		return nil
 	})
 	if err != nil {
-		t.Error(err)
+		t.Fatal(err)
```
samvantran: You may want to [read this thread](https://github.com/mesosphere/kudo-cassandra-operator/pull/3#discussion_r339744459) as there are alternate solutions for handling Errors

tests/security_test.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	spark := utils.SparkOperatorInstallation{
+		Namespace:            namespace,
+		SkipNamespaceCleanUp: true,
+		Params: map[string]string{
+			"createOperatorServiceAccount": "false",
+			"operatorServiceAccountName":   operatorSA,
+			"createSparkServiceAccount":    "true",
+			"sparkServiceAccountName":      driverSA,
```
samvantran: seems like a lot of this test code is duplicated? may I suggest a helper function that enables toggling b/w `true` `false` for the spark/operator service accounts... like you did with `testServiceAccounts`?

tests/security_test.go
rpalaznik: Makes sense. I refactored the cases into sub-tests.

tests/security_test.go
```diff
@@ -0,0 +1,146 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			err := runServiceAccountTestCase(tc)
+			if err != nil {
+				t.Errorf("Test case: %v\nfailed with error: %s", tc, err)
+			}
+		})
+	}
+}
+
+func runServiceAccountTestCase(tc serviceAccountTestCase) error {
```
VladimirKravtsov: The method still looks too long on my taste. But overall the PR's code looks really great now! Many thanks for the refactorings, @rpalaznik!

PR #46, opened 2019-10-30
[DCOS-59796]: Integration test for Operator tenancy

alembiewski:
Resolves [DCOS-59796](https://jira.mesosphere.com/browse/DCOS-59796)

Add a set of integration tests to verify operator multi-tenancy by checking the main components (`Service`, `ServiceAccount`, `Role`, `ClusterRole`, `ClusterRoleBinding`, `CRD`) and by running a simple workload on each installation.

Tests cover the following scenarios:
- install two operator instances in two namespaces
- install two operator instances with the same name in two namespaces
- install two operator instances in a single namespace
- CRDs deletion

tested locally on `minikube` and by running the build in CI

### DIFF
--- a/tests/templates/spark-pi.yaml
+++ b/tests/templates/spark-pi.yaml
@@ -1,0 +1,30 @@
+apiVersion: "sparkoperator.k8s.io/v1beta2"
+kind: SparkApplication
+metadata:
+  name: {{ .Name }}
+  namespace: {{ .Namespace }}
+spec:
+  type: Scala
+  mode: cluster
+  image: {{ .Image }}
+  imagePullPolicy: Always
+  mainClass: org.apache.spark.examples.SparkPi
+  mainApplicationFile: "local:///opt/spark/examples/jars/spark-examples_2.11-{{ .SparkVersion }}.jar"
+  sparkConf:
+    "spark.scheduler.maxRegisteredResourcesWaitingTime": "2400s"
+    "spark.scheduler.minRegisteredResourcesRatio": "1.0"
+  sparkVersion: {{ .SparkVersion }}
+  restartPolicy:
+    type: Never
+  driver:
+    cores: 1
+    memory: "512m"
+    labels:
+      version: {{ .SparkVersion }}
+    serviceAccount: spark-driver
+  executor:
+    cores: 1
+    instances: 1
+    memory: "512m"
+    labels:
+      version: {{ .SparkVersion }}
--- a/tests/templates/spark-pi.yaml
+++ b/tests/templates/spark-pi.yaml
@@ -21,7 +21,7 @@ spec:
     memory: "512m"
     labels:
       version: {{ .SparkVersion }}
-    serviceAccount: spark-driver
+    serviceAccount: {{ .ServiceAccount }}
   executor:
     cores: 1
     instances: 1
--- a/tests/tenancy_test.go
+++ b/tests/tenancy_test.go
@@ -11,9 +11,9 @@ import (
 )
 
 func TestTenancyTwoOperatorsDifferentNamespaces(t *testing.T) {
-	operators := operatorBuilder(2, true, true)
+	operators := operatorBuilder(1, true, true)
 	for _, operator := range operators {
-		err := operator.InstallSparkOperator(true)
+		err := operator.InstallSparkOperator()
 		assert.NilError(t, err)
 	}
 
@@ -37,8 +37,11 @@ func TestTenancyTwoOperatorsDifferentNamespaces(t *testing.T) {
 
 func TestTenancyTwoOperatorsSingleNamespace(t *testing.T) {
 	operators := operatorBuilder(2, false, true)
-	for _, operator := range operators {
-		err := operator.InstallSparkOperator(false)
+	for i, operator := range operators {
+		if i > 0 {
+			operator.SkipNamespaceCleanUp = true
+		}
+		err := operator.InstallSparkOperator()
 		assert.NilError(t, err)
 		defer operator.CleanUp()
 	}
@@ -54,7 +57,7 @@ func TestTenancyTwoOperatorsSingleNamespace(t *testing.T) {
 func TestTenancyTwoOperatorsSameNameDifferentNamespaces(t *testing.T) {
 	operators := operatorBuilder(2, true, false)
 	for _, operator := range operators {
-		err := operator.InstallSparkOperator(true)
+		err := operator.InstallSparkOperator()
 		assert.NilError(t, err)
 		defer operator.CleanUp()
 	}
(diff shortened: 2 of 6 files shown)


tests/templates/spark-pi.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: "sparkoperator.k8s.io/v1beta2"
```
akirillov: can we use mock task runner job as default here? nothing prevents us from using Spark Pi but MTR is deterministic with respect to execution time and can be launched with some small sleep time to avoid long waits.

alembiewski: good point, but I would suggest to keep using `SparkPi` as a test job because it is faster in comparison with `MockTaskRunner`. In case with the latter, a job needs to fetch an external artifact (> 60 Mb), and that will increase test times, and for the former, the jar is already present inside the image. I've compared two examples locally, `SparkPi` was a way faster then MTR, started with the small arguments' values (1 task 1 second).

akirillov: good point, thanks. we download MTR anyway for testing the killing of running application where we need more control over the duration to avoid flakiness due to timing issues - so the download time is already there.

We also user Spark Pi in DCOS Spark tests to verify Driver output - so leaving it up to you, just wanted to keep the number of simple apps (not targeting any specific corner-cases like shuffle) low.

alembiewski: > we download MTR anyway for testing the killing of running application where we need more control over the duration to avoid flakiness due to timing issues - so the download time is already there.

yes, but from my observations during local testing, it won't be reused across different operator instances even in the same namespace from one test run to another.

alembiewski:
Thanks for the review @akirillov! 
CI check has passed, but I would like to postpone merging this PR before #45 is merged because both PRs contain similar changes related to conditional cleanup in `InstallSparkOperator` method, and I like @rpalaznik's approach of handling that, so I'm going to rebase my branch on top of this changes.

tests/tenancy_test.go
```diff
@@ -0,0 +1,150 @@
...
+		for _, role := range roles {
+			if strings.Contains(role, "%s") {
+				role = fmt.Sprintf(role, operator.InstanceName)
+			}
+			log.Infof("Checking Role \"%s\" is created in namespace \"%s\" for \"%s\"",
+				role, operator.Namespace, operator.InstanceName)
+			result, err := operator.K8sClients.RbacV1().Roles(operator.Namespace).Get(role, v1.GetOptions{})
+			assert.NilError(t, err)
+			instanceLabel, present := result.Labels["kudo.dev/instance"]
+			if present {
```
VladimirKravtsov: Why if the label is not present here, the situation is ignored?

tests/tenancy_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+}
+
+func verifyComponents(t *testing.T, operators []*utils.SparkOperatorInstallation) {
+	serviceAccounts := []string{"spark-operator-service-account", "spark-service-account"}
+	services := []string{"spark-webhook", "spark-operator-metrics"}
+	roles := []string{"spark-driver-role", "%s-spark-role"}
+
+	for _, operator := range operators {
+		for _, service := range services {
```
VladimirKravtsov: I'd suggest wrapping each following subtests (verify Services, ServiceAccounts, Roles etc.) into `t.Run`, so making them atomical. It will more comply with the unit testing philosophy.

The example of how it could looks like
```suggestion
for _, operator := range operators {
        t.Run("TestServices", func(t *testing.T) {
		for _, service := range services {
					serviceName := fmt.Sprint(operator.InstanceName, "-", service)
 			log.Infof("Checking Service \"%s\" is created in namespace \"%s\" for \"%s\"", serviceName,
 				operator.Namespace, operator.InstanceName)
 			result, err := operator.K8sClients.CoreV1().Services(operator.Namespace).Get(fmt.Sprint(serviceName), v1.GetOptions{})
 			assert.NilError(t, err)
 			assert.Equal(t, result.Labels["kudo.dev/instance"], operator.InstanceName)
 		}
 	}(t)

```

tests/tenancy_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+func TestTenancyTwoOperatorsSameNameDifferentNamespaces(t *testing.T) {
+	operators := operatorBuilder(2, true, false)
+	for _, operator := range operators {
+		err := operator.InstallSparkOperator(true)
+		assert.NilError(t, err)
+	}
+
+	verifyComponents(t, operators)
+	verifyWorkloads(t, operators)
```
VladimirKravtsov: Here and in the above method:
```suggestion
t.Run("TestWorkloads", func(t *testing.T) {
	verifyWorkloads(t, operators)
})
```

tests/tenancy_test.go
VladimirKravtsov: Also, the tests which are wrapped into `t.Run` could be run in parallel. The optimization can bring faster execution and atomical the tests fails.

tests/tenancy_test.go
alembiewski: yes, because in this particular case not all the `Role` resources contain this label that we are verifying, therefore we can't explicitly check to which instance it belongs to. But we are still checking the role exists it the operator's namespace, so this test is valid.

tests/templates/spark-pi.yaml
akirillov: that's right. Spark pulls application jar every time a job is submitted, good point.

alembiewski:
Thanks for the review, @VladimirKravtsov!
I decomposed the tests based on the approach you suggested,  and it looks good from my perspective: tests became more granular and self-descriptive. Here is how it looks like in CI:
https://teamcity.mesosphere.io/viewLog.html?buildTypeId=Frameworks_DataServices_Kudo_Spark_Pr_SparkPrKonvoyKudoV073&buildId=2400816&tab=testsInfo&pager.currentPage=1

tests/tenancy_test.go
VladimirKravtsov: ok, got it

akirillov:
just as a side-note, I think with the new `t.Run` approach the test cases reports became fuzzier. The semantics of the test cases here is the following: "All the components of the Operator are installed under these conditions". In the test report, it breaks down to 6 independent tests per each condition which makes it hard to reason about what was the main scenario.

CI now reports 45 independent tests which I believe does not reflect the actual use cases. I'm not sure it's worth the parallel optimization:
CI test report sample

alembiewski:
@akirillov, 
I agree, that we probably don't need such detailed report about all of the components, so I've made the test-suite less granular:
image

PR #49, opened 2019-11-02
Scale test scripts updates to support Terasort benchmark, improved S3 write time

akirillov:
Resolves [DCOS-60318: Improve gensort job to avoid high latencies while writing to S3](https://jira.mesosphere.com/browse/DCOS-60318)

Originally started as gensort write time improvement which is achieved via committer configuration property:
```
"spark.hadoop.mapreduce.fileoutputcommitter.algorithm.version": "2"
```
This PR includes refactoring to separate scale tests from the sort benchmark and introduces changes to support ongoing work in https://github.com/mesosphere/spark-build/pull/559 and https://github.com/mesosphere/spark-build/pull/560

These changes are needed to support new scale tests objectives and run the workloads separately.

* by running `gensort` job on a cluster with 40 executors and verifying 1TiB dataset is generated in under 1 hour
* by running all the scripts against private cluster to test changes from https://github.com/mesosphere/spark-build/pull/559 and https://github.com/mesosphere/spark-build/pull/560

### DIFF
--- a/scale-tests/scripts/terasort.sh
+++ b/scale-tests/scripts/terasort.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 #
-# This script runs a dataset generator for sorting benchmark and requires Spark Operator to be installed
+# This script submits a sorting application and requires Spark Operator to be installed.
+# Dataset should be generated upfront.
 
 set -ex
 SCRIPT_DIR="$(realpath "$(dirname "$0")")"
@@ -14,7 +15,7 @@ SERVICE_ACCOUNT_NAME=${SERVICE_ACCOUNT_NAME:-spark-service-account}
 
 if [[ $# -lt 2 ]]; then
   echo "Usage:" >&2
-  echo "  $0 <namespace> <target s3 path>" >&2
+  echo "  $0 <namespace> <S3 source path> <S3 target path>" >&2
   exit 1
 fi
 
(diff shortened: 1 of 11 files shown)


scale-tests/scripts/terasort.sh
```diff
@@ -0,0 +1,33 @@
...
+. ${SCRIPT_DIR}/aws_credentials.sh
+
+cat ${TEMPLATES_DIR}/sort-application.tmpl \
+  | sed "s|AWS_ACCESS_KEY_ID|${AWS_ACCESS_KEY_ID:-}|" \
+  | sed "s|AWS_SECRET_ACCESS_KEY|${AWS_SECRET_ACCESS_KEY:-}|" \
+  | sed "s|AWS_SESSION_TOKEN|${AWS_SESSION_TOKEN:-}|" \
+  | sed "s|S3_ENDPOINT|${S3_ENDPOINT}|" \
+  | sed "s|NUM_EXECUTORS|${NUM_EXECUTORS}|" \
+  | sed "s|SOURCE_PATH|${2:-}|" \
+  | sed "s|TARGET_PATH|${3:-}|" | kubectl --namespace "${NAMESPACE}" apply -f -
```
VladimirKravtsov: Here is a missed new line and maybe move the `| kubectl --namespace "${NAMESPACE}" apply -f -` on new line would be better?
```suggestion
  | sed "s|TARGET_PATH|${3:-}|" 
  | kubectl --namespace "${NAMESPACE}" apply -f -
  
```

scale-tests/templates/gensort-application.tmpl
```diff
@@ -39,7 +40,7 @@ spec:
     memory: "1g"
     labels:
       version: 2.4.3
-    serviceAccount: spark-driver
+    serviceAccount: spark-service-account
```
farhan5900: Should service account name come from `gensort.sh` using environment variable? As user can change it by exporting `SERVICE_ACCOUNT_NAME` environment variable.
```suggestion
    serviceAccount: SERVICE_ACCOUNT_NAME
```

scale-tests/templates/sort-application.tmpl
```diff
@@ -1,39 +1,43 @@
...
     type: Never
   driver:
     cores: 2
-    memory: "2g"
+    memory: "6g"
     labels:
       version: 2.4.3
       metrics-exposed: "true"
-    serviceAccount: spark-driver
+    serviceAccount: spark-service-account
```
farhan5900: Similar as `gensort` template.

scale-tests/scripts/terasort.sh
akirillov: sounds good, thanks for pointing that out. updated here and in other similar scripts

scale-tests/templates/gensort-application.tmpl
akirillov: Although we don't need a configurable service account for scale tests I updated the code to support the change you've suggested. The original reasoning was about having a hard-coded  static service account for all scale-test applications and create this account in `install.sh` script. But now it's templated as per your suggestion.

scale-tests/templates/sort-application.tmpl
akirillov: fixed

PR #59, opened 2019-11-11
[DCOS-60858] [Spark Operator] Test mounting secrets

farhan5900:
Resolves [DCOS-60858](https://jira.mesosphere.com/browse/DCOS-60858)

Implements integration tests for environment based secret as well as file based secret. It creates a secret and for the case of file based test, it mounts secret as file and check mounted path in job description. For the case of env based test, it set an environment variable by referencing it to a secret key and check for the environment variable in job description.

To verify File based as well as env based secret working.

Ran `make test` locally.

### DIFF
--- a/tests/security_test.go
+++ b/tests/security_test.go
@@ -344,8 +344,11 @@ func runSecretTest(secretName string, secretPath string, secretKey string) (stri
 		return "", err
 	}
 
-	secretData := map[string]string{
-		secretKey: "secretValue",
+	secretData := make(map[string]string)
+	if secretKey != "" {
+		secretData[secretKey] = "secretValue"
+	} else {
+		secretData["secretKey"] = "secretValue"
 	}
 
 	err = utils.CreateSecret(client, secretName, spark.Namespace, secretData)
--- a/tests/security_test.go
+++ b/tests/security_test.go
@@ -298,50 +298,34 @@ func runTestCase(tc securityTestCase) error {
 	return err
 }
 
-func TestEnvBasedSecret(t *testing.T) {
-	secretName := "env-based-secret"
-	secretKey := "secretKey"
-	jobDescription, err := runSecretTest(secretName, "", secretKey)
+func TestEnvBasedSecrets(t *testing.T) {
+	err := runSecretTest("env-based-secret", "", "secretKey", "set to the key 'secretKey' in secret 'env-based-secret'")
 
 	if err != nil {
 		t.Error(err.Error())
-	}
-
-	if strings.Contains(jobDescription, fmt.Sprintf("set to the key '%s' in secret '%s'", secretKey, secretName)) {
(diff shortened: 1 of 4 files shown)


tests/security_test.go
```diff
@@ -295,3 +297,96 @@ func runTestCase(tc securityTestCase) error {
...
+	}
+
+	if strings.Contains(jobDescription, fmt.Sprintf("set to the key '%s' in secret '%s'", secretKey, secretName)) {
+		log.Infof("Successfully set environment variable to the key '%s' in secret '%s'", secretKey, secretName)
+	} else {
+		t.Errorf("Unnable to set environment variable to the key '%s' in secret '%s'", secretKey, secretName)
+	}
+}
+
+func TestFileBasedSecrets(t *testing.T) {
```
VladimirKravtsov: These two methods(`TestEnvBasedSecret` and `TestFileBasedSecrets`)  are looking very similar. maybe it makes sense to extract the logic into another one method and invoke it with different parameters depending on test case.  
WDYT, @farhan5900 ?

farhan5900: Right these two methods are almost same, that's why we have `runSecretTest` method containing all common logic. We can extract remaining logic as well. I will try to do that.

tests/security_test.go
```diff
@@ -295,3 +297,93 @@ func runTestCase(tc securityTestCase) error {
...
+func runSecretTest(secretName string, secretPath string, secretKey string, expectedSecret string) error {
+	spark := utils.SparkOperatorInstallation{}
+	err := spark.InstallSparkOperator()
+	defer spark.CleanUp()
+
+	if err != nil {
+		return err
+	}
+
+	client, err := utils.GetK8sClientSet()
```
rpalaznik: After calling `spark.InstallSparkOperator()`, the `spark` struct will already have a client set ready to use as `spark.K8sClients` which you can pass to `CreateSecret` function

tests/security_test.go
```diff
@@ -295,3 +297,93 @@ func runTestCase(tc securityTestCase) error {
...
+	err = spark.WaitUntilSucceeded(job)
+	if err != nil {
+		return err
+	}
+
+	jobDescription, err := utils.Kubectl(
+		"describe",
+		"pod",
+		"--namespace="+spark.Namespace,
+		jobName+"-driver",
```
rpalaznik: There is a private util function that does the same https://github.com/mesosphere/kudo-spark-operator/blob/ac69995410810bd83afcff3bf3b3f4c7b4b2021a/tests/utils/job.go#L80-L82  Consider making it public and using it here. It's unlikely, but if driver pod naming changes, it's better to have only one place to fix.

tests/security_test.go
farhan5900: Thanks @rpalaznik for pointing it out. Changed it.

tests/security_test.go
farhan5900: Yeah it is better way. Made the method public.

PR #60, opened 2019-11-13
Fix mutating admission webhook

alembiewski:
### What changes were proposed in this pull request?
This PR contains fixes for mutation admission webhook, which used to customize Spark driver and executor pods based on the specification in `SparkApplication` objects. 

After the fix is accepted and verified, it should be applied to https://github.com/kudobuilder/operators/pull/118 as well.

### Why are the changes needed?
While working on [DCOS-60865: Test pod affinity and tolerations support](https://jira.mesosphere.com/browse/DCOS-60865), it was discovered, that aforementioned features didn't work as expected, even though webhook was enabled by default and service was installed. Operator's description showed the following warning:
```
➜ k describe pod test-instance-5869fb6d47-7tqw5 -n kudo-spark-operator-testing
Warning  FailedMount  66s (x3 over 68s)  kubelet, minikube  MountVolume.SetUp failed for volume "webhook-certs" : secret "spark-webhook-certs" not found
```
But `spark-webhook-certs` was actually present:
```
➜ k get secrets spark-webhook-certs -n kudo-spark-operator-testing        
NAME                  TYPE     DATA   AGE
spark-webhook-certs   Opaque   4      4m19s
```
If we look at  the `AGE` columns for each object, we can see the operator instance `test-instance-5869fb6d47-7tqw5` was created before `spark-webhook-certs` secret:

image

Also, `-webhook-svc-name` was changed to adhere this declaration: 
https://github.com/mesosphere/kudo-spark-operator/blob/2d7765b2569e551d53ed5c14ac93c232aabc52e0/kudo-operator/operator/templates/webhook-init-job.yaml#L23

After those fixes were applied, `ownerReferences`, `affinity` and `tolerations` appeared in `Pod`'s spec and worked as expected:
```
➜  kudo-spark-operator git:(test-affinity-and-tolerations) ✗ k get pod spark-pi-driver -o yaml -n kudo-spark-operator-testing 
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: "2019-11-13T11:07:38Z"
  labels:
    spark-app-selector: spark-f3c0ec9c38ac469cb571a9f62cd98702
    spark-role: driver
    sparkoperator.k8s.io/app-name: spark-pi
    sparkoperator.k8s.io/launched-by-spark-operator: "true"
    sparkoperator.k8s.io/submission-id: c0143db9-ce40-4d45-a184-bc6e6ef7198f
    version: 2.4.3
  name: spark-pi-driver
[...]

PR #61, opened 2019-11-13
DCOS-60860 - Test host network support

rpalaznik:
Resolves [DCOS-60860](https://jira.mesosphere.com/browse/DCOS-60860)
- Adds a test for SparkApplication `spec.hostNetwork` parameter behavior according to the [user guide](https://github.com/mesosphere/spark-on-k8s-operator/blob/master/docs/user-guide.md#host-network)

Please note, that this PR temporarily includes changes from #60 as well, to prevent CI builds from failing. Once #60 is merged I'll rebase this PR to make the diff cleaner.

With `make test` in a k8s cluster in AWS.

### DIFF
--- a/tests/network_test.go
+++ b/tests/network_test.go
@@ -3,7 +3,6 @@ package tests
 import (
 	"errors"
 	"fmt"
-	"github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/apis/sparkoperator.k8s.io/v1beta2"
 	"github.com/mesosphere/kudo-spark-operator/tests/utils"
 	log "github.com/sirupsen/logrus"
 	v12 "k8s.io/api/core/v1"
@@ -45,30 +44,10 @@ func TestHostNetworkPropagation(t *testing.T) {
 			},
 		}
 
-		expectedExecutorCount := 1
-
 		// Submit the job and wait for it to start
-		err = spark.SubmitJob(&job)
+		err = spark.SubmitAndWaitForExecutors(&job)
 		if err != nil {
 			t.Fatal(err)
-		}
-		err = spark.WaitForJobState(job, v1beta2.RunningState)
-		if err != nil {
-			t.Fatal(err)
-		}
-
-		// Wait for correct number of executors to show up
-		err = utils.Retry(func() error {
-			executors, err := spark.GetExecutorState(job)
-			if err != nil {
-				return err
-			} else if len(executors) != expectedExecutorCount {
-				return errors.New(fmt.Sprintf("The number of executors is %d, but %d is expected", len(executors), expectedExecutorCount))
-			}
-			return nil
-		})
-		if err != nil {
-			t.Error(err)
 		}
 
 		// Verify driver pod hostNetwork and dnsPolicy values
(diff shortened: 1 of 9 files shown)


tests/network_test.go
```diff
@@ -0,0 +1,105 @@
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		log.Infof("Driver spec.hostNetwork: %v", driver.Spec.HostNetwork)
+		log.Infof("Driver spec.dnspolicy: %v", driver.Spec.DNSPolicy)
+		if driver.Spec.HostNetwork != tc.driverHN {
+			t.Fatal(fmt.Sprintf("Unexpected hostNetwork value for driver %v: %s. Should be %v", driver.Spec.HostNetwork, driver.Name, tc.driverHN))
+		} else if tc.driverHN && driver.Spec.DNSPolicy != v12.DNSClusterFirstWithHostNet {
+			t.Fatal(fmt.Sprintf("Expected driver pod DNS policy to be \"dnsClusterFirstWithHostNet\", but it's %s", driver.Spec.DNSPolicy))
+		}
```
alembiewski: it looks like this check could be generalized and extracted to a separate method, which could accept `pod Pod` and pod's type as arguments. WDYT?

tests/network_test.go
```diff
@@ -0,0 +1,105 @@
...
+		// Verify executor pods hostNetwork and dnsPolicy values
+		executors, err := spark.ExecutorPods(job)
+		if err != nil {
+			t.Fatal(err)
+		}
+		for _, executor := range executors {
+			log.Infof("Executor %s spec.hostNetwork: %v", executor.Name, executor.Spec.HostNetwork)
+			log.Infof("Executor %s spec.dnsPolicy: %v", executor.Name, executor.Spec.DNSPolicy)
+			if executor.Spec.HostNetwork != tc.executorHN {
+				t.Fatal(fmt.Sprintf("Unexpected hostNetwork value for driver %v: %s. Should be %v", executor.Spec.HostNetwork, executor.Name, tc.executorHN))
```
alembiewski:
```suggestion
				t.Fatal(fmt.Sprintf("Unexpected hostNetwork value for executor %v: %s. Should be %v", executor.Spec.HostNetwork, executor.Name, tc.executorHN))
```

tests/network_test.go
```diff
@@ -0,0 +1,105 @@
...
+			} else if tc.executorHN && executor.Spec.DNSPolicy != v12.DNSClusterFirstWithHostNet {
+				t.Fatal(fmt.Sprintf("Expected executor pod DNS policy to be \"dnsClusterFirstWithHostNet\", but it's %s", executor.Spec.DNSPolicy))
+			}
```
alembiewski: ditto

tests/network_test.go
```diff
@@ -0,0 +1,105 @@
...
+			if err != nil {
+				return err
+			} else if len(executors) != expectedExecutorCount {
+				return errors.New(fmt.Sprintf("The number of executors is %d, but %d is expected", len(executors), expectedExecutorCount))
+			}
+			return nil
+		})
+		if err != nil {
+			t.Error(err)
+		}
```
akirillov: this part looks generalizable and applicable to all the tests where jobs are submitted. Can we extract this logic in some generic `SubmitAndWait` or alike?

tests/network_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+		// Verify driver pod hostNetwork and dnsPolicy values
+		driver, err := spark.DriverPod(job)
+		if err != nil {
+			t.Fatal(err)
+		}
+		log.Infof("Driver spec.hostNetwork: %v", driver.Spec.HostNetwork)
+		log.Infof("Driver spec.dnspolicy: %v", driver.Spec.DNSPolicy)
+		if driver.Spec.HostNetwork != tc.driverHN {
```
akirillov: do you think pod spec check is sufficient compared to the actual IP address check?

tests/network_test.go
rpalaznik: done

tests/network_test.go
rpalaznik: Make sense. Created a method for it.

tests/network_test.go
rpalaznik: Added IP check as well

PR #65, opened 2019-11-16
[DCOS-60866] Test pod security context

alembiewski:
Resolves [DCOS-60866: Test pod security context](https://jira.mesosphere.com/browse/DCOS-60866)

This PR introduces a test for pod security context as well as a few fixes in Spark base image Dockerfile, which allow running Spark jobs as a non-root user.

Due to insufficient permissions, Spark application failed with errors when running as a `nobody` user.
Pod spec:
```
securityContext:
      runAsUser: 65534
      runAsGroup: 65534
      runAsNonRoot: true
```
Exception in Driver's log:
```
Exception in thread "main" java.lang.NoClassDefFoundError: io/fabric8/kubernetes/client/KubernetesClient
```
```bash
bash-4.4# ls -l /opt/spark/jars/ | grep kubernetes
-rw-------    1 root     root        616644 Aug 23 16:48 kubernetes-client-4.4.2.jar
-rw-r--r--    1 root     root       9386214 Oct 18 01:15 kubernetes-model-4.1.2.jar
-rw-r--r--    1 root     root          3817 Oct 18 01:15 kubernetes-model-common-4.1.2.jar
-rw-r--r--    1 root     root        538176 Oct 18 01:15 spark-kubernetes_2.11-2.4.3.jar
```
Also, write permissions were added for `/opt/spark/work-dir` to allow saving job-related files. 

- by running a test on a private cluster
- integration tests from this repo

### DIFF
--- a/tests/security_test.go
+++ b/tests/security_test.go
@@ -3,6 +3,7 @@ package tests
 import (
 	"errors"
 	"fmt"
+	"gotest.tools/assert"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


tests/security_test.go
```diff
@@ -382,3 +383,39 @@ func runSecretTest(secretName string, secretPath string, secretKey string, expec
 
 	return nil
 }
+
+func TestPodSecurityContext(t *testing.T) {
+
+	// run driver and executor as user "nobody"
+	const uid = "65534"
+	const gid = "65534"
```
akirillov: just wondering, can we use user and group names instead of UIDs?

tests/security_test.go
```diff
@@ -382,3 +383,39 @@ func runSecretTest(secretName string, secretPath string, secretKey string, expec
...
+
+	err = spark.SubmitJob(&job)
+	assert.NilError(t, err)
+
+	// verify job completed successfully
+	err = spark.WaitUntilSucceeded(job)
+	assert.NilError(t, err)
+
+	// verify uid propagated to the pod
+	logContains, err := spark.DriverLogContains(job, fmt.Sprintf("myuid=%s", uid))
```
akirillov: is there a way to verify container's user via metadata instead of logs?

tests/security_test.go
alembiewski: Looks like it's not possible. As per [k8s api reference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#securitycontext-v1-core), those fields must of type `integer`.

tests/security_test.go
alembiewski: yes, sounds reasonable, thanks! Added check of driver and executor pod specs in addition to driver's logs verification

PR #68, opened 2019-11-29
[DCOS-61531] Use separate Docker repositories for developer and release builds

alembiewski:
Resolves [DCOS-61531](https://jira.mesosphere.com/browse/DCOS-61531).

Changes:
- Images, produced during CI builds are now pushed to `*-dev` repository
- Added new `make` target for release preparation.

Separation by artifact type will bring the following benefits:
- isolation - production and dev artifacts won't be mixed together in a single repo;
- image versioning/tagging - different policies could be used for image versioning, e.g. checksums could only be used for CI images, while production image versions could be more human-readable.

Tests from this repo.


Makefile
```diff
@@ -34,6 +34,9 @@ DOCKER_BUILDER_IMAGE_DIR ?= $(ROOT_DIR)/images/builder
 DOCKER_BUILDER_IMAGE_TAG ?= $(call files_checksum,$(DOCKER_BUILDER_IMAGE_DIR))
 DOCKER_BUILDER_IMAGE_FULL_NAME ?= $(DOCKER_REPO_NAME)/$(DOCKER_BUILDER_IMAGE_NAME):$(DOCKER_BUILDER_IMAGE_TAG)
 
+OPERATOR_IMAGE_RELEASE_NAME ?= kudo-spark-operator
+SPARK_IMAGE_RELEASE_NAME ?= spark
```
akirillov: given the nature of these variables, shall we default them to the full tags? e.g.
```
OPERATOR_IMAGE_RELEASE_TAG ?= $(DOCKER_REPO_NAME)/kudo-spark-operator:latest
SPARK_IMAGE_RELEASE_TAG ?= $(DOCKER_REPO_NAME)/spark:2.4.3-k8s-latest
```

And in the release job we will specify these tags fully. Another option would be to pass release tags only while keeping org/repo names static.

Makefile
```diff
@@ -19,12 +19,12 @@ KUBECONFIG ?= $(ROOT_DIR)/admin.conf
...
 DOCKER_REPO_NAME ?= mesosphere
 
-SPARK_IMAGE_NAME ?= spark
+SPARK_IMAGE_NAME ?= spark-dev
 SPARK_IMAGE_DIR ?= $(ROOT_DIR)/images/spark
 SPARK_IMAGE_TAG ?= $(call files_checksum,$(SPARK_IMAGE_DIR))
 SPARK_IMAGE_FULL_NAME ?= $(DOCKER_REPO_NAME)/$(SPARK_IMAGE_NAME):$(SPARK_IMAGE_TAG)
 
-export OPERATOR_IMAGE_NAME ?= kudo-spark-operator
+export OPERATOR_IMAGE_NAME ?= kudo-spark-operator-dev
```
akirillov: while we're here, can we refactor the naming a bit because `name` doesn't fit DockerHub terminology?
Suggestions:
* `DOCKER_REPO_NAME` -> `DOCKER_ORG`
* `SPARK_IMAGE_NAME ` -> `SPARK_DOCKER_REPO`
* `OPERATOR_IMAGE_NAME ` -> `OPERATOR_DOCKER_REPO`

Probably, an even better option would be to get rid of `DOCKER_REPO_NAME` and specify the org as a part of image repo names. WDYT?

alembiewski: > Probably, an even better option would be to get rid of DOCKER_REPO_NAME and specify the org as a part of image repo names. WDYT?

Yes, probably it doesn't make sense to have it as a variable, taking into account the separation between dev and release repositories. I believe the main idea of parametrizing that was for easy switching to personal docker hub account for testing purposes.

Makefile
```diff
@@ -132,6 +135,14 @@ test:
 install:
 	OPERATOR_IMAGE_NAME=$(DOCKER_REPO_NAME)/$(OPERATOR_IMAGE_NAME) OPERATOR_VERSION=$(OPERATOR_VERSION) $(SCRIPTS_DIR)/install_operator.sh
 
+.PHONY: release
+release: SPARK_IMAGE_NAME = $(SPARK_IMAGE_RELEASE_NAME)
+release: OPERATOR_IMAGE_NAME = $(OPERATOR_IMAGE_RELEASE_NAME)
+release: docker-push
```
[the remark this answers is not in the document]
alembiewski: good point, thanks! reworked that as suggested

Makefile
alembiewski: I assume org and repo names will not change much, so I would rather go with the second option.

alembiewski:
Thanks for the review @akirillov! Updated PR, PTAL.
Also, I've started implementing the [release job](https://teamcity.mesosphere.io/admin/editBuild.html?id=buildType:Frameworks_DataServices_Kudo_Spark_Tools_Release), it would be nice if you take a look at it as well. The job is missing `docker login`, will add it later, going to make some `dry-runs` of the job to verify the configuration.
As a side-note, `kudo-spark-operator-dev` Docker repo was created as a private repo, so calls to the registry are not working properly, as a result, operator image gets rebuilt every-time. I've already created the request to HD to make it public.

PR #70, opened 2019-12-10
[DCOS-61620] Provide and test SparkR support

alembiewski:
Resolves [DCOS-61620](https://jira.mesosphere.com/browse/DCOS-61620)
This PR adds SparkR support to base Spark image as well as an integration test for exercising R API.  

Add the ability to run SparkR applications on Kudo Spark Operator by updating the base Spark image. 
Also, an integration test was added to the test suite. The test submits SparkR application, which makes a prediction based on Alternating Least Squares algorithm. 

By running the test locally on minikube;
By running tests from this repo in CI.


alembiewski:
@akirillov, good catch, thank you!
Indeed, we need to support `sparkr` profile in our build job as well, but it will require R package to be installed on a build environment, so I  am going to tweak the build job a bit (by introducing `spark-build` image), build a distribution and verify it using our test suite.

PR #73, opened 2019-12-12
Update CODEOWNERS

alembiewski:
This PR actualizes CODEOWNERS file according to the current team structure.  

To prevent automatic review requests to team members, who are not currently involved in the project.


alembiewski:
Thanks @samvantran :disappointed: 
@akirillov, updated the description, thanks for pointing it out! Looks like CI check has failed: Spark History test fails with `Access Denied` error while trying to create an S3 bucket. Could be related to the recent account changes. Will create a separate ticket for that matter.

PR #74, opened 2019-12-17
[DCOS-61938] Integration test for HA controller deployment.

alembiewski:
Resolves [DCOS-61938: Integration test and documentation for HA controller deployment](https://jira.mesosphere.com/browse/DCOS-61938).
This PR introduces a new test suite for HA deployments.

In order to verify HA deployment, the following tests were introduced:
- test HA deployment and leader election parameters propagation
- test leader election process
- test leader re-election when the current leader is down while running SparkApplication and verify the application state updates.

- test from this repo


tests/ha_test.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	log.Infof("deleting current leader pod \"%s\"", leaderElectionRecord.HolderIdentity)
+	err = utils.DeleteResource(operator.Namespace, "pod", leaderElectionRecord.HolderIdentity)
+
+	// check re-election
+	if err := utils.RetryWithTimeout(electionRecordRetryTimeout, electionRecordRetryInterval, func() error {
+		if newLeaderElectionRecord, err := getLeaderElectionRecord(operator); err != nil {
+			return err
+		} else if newLeaderElectionRecord.HolderIdentity == leaderElectionRecord.HolderIdentity {
```
akirillov: is there a possibility that the last leader can be elected again? this can expose the test to timing issues and make it flaky. Does it make sense to assign lower values to election-related time settings in parameters?

tests/ha_test.go
```diff
@@ -0,0 +1,147 @@
...
+			return errors.New("Waiting for the new leader to be elected")
+		} else {
+			log.Info("New leader found: ", newLeaderElectionRecord.HolderIdentity)
+		}
+		return nil
+	}); err != nil {
+		suite.FailNow(err.Error())
+	}
+
+	suite.NoError(operator.WaitForJobState(mockTaskRunner, v1beta2.CompletedState))
```
akirillov: Once Spark job is submitted in cluster mode,  technically, it doesn't need a client machine/pod to be up anymore and can complete independently. What would be nice to verify is that the original leader submit the job and the newly elected leader handles job status update when it completes. Otherwise, in the current implementation, there's no link between failover and status update handling. 

How complicated would it be to implement that testing scenario?

As an alternative, we can simplify it a bit and submit two independent jobs before and after reelection and verify they're handled by different controller instances.

tests/ha_test.go
```diff
@@ -0,0 +1,147 @@
...
+		Template: "spark-mock-task-runner-job.yaml",
+		Params: map[string]interface{}{
+			"args": []string{"1", "30"},
+		},
+	}
+	if err := operator.SubmitJob(&mockTaskRunner); err != nil {
+		suite.FailNow(err.Error())
+	}
+	// wait until the application is running
+	if err := operator.WaitForJobState(mockTaskRunner, v1beta2.RunningState); err != nil {
```
akirillov: Does it make sense to check the current leader and other controllers logs to verify that the submission is handled by the leader only? In the current implementation, the job state is checked bypassing the controller and doesn't verify that.

tests/ha_test.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func (suite *HighAvailabilityTestSuite) SetupSuite() {
+	suite.leaderElectionParams = map[string]string{
+		"replicas":                    "3",
+		"enableLeaderElection":        "true",
+		"leaderElectionLockName":      "lock",
+		"leaderElectionLeaseDuration": "15s",
+		"leaderElectionRenewDeadline": "10s",
+		"leaderElectionRetryPeriod":   "3s",
```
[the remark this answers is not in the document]
alembiewski: Yes, I believe it does. One drawback of using `struct` is that we need to convert it to a `map[string]string` before passing it to `SparkOperatorInstallation`, so we need to write an additional method that converts `struct` to `map`. I will try to work this out.

tests/ha_test.go
alembiewski: I would say no because we delete the leader pod and the new pod is created instead by `ReplicaSet`. This condition is required to check that the leader election record is updated with the newly elected leader and force the method to retry if `HolderIdentity`  isn't changed.

tests/ha_test.go
alembiewski: > Once Spark job is submitted in cluster mode, technically, it doesn't need a client machine/pod to be up anymore and can complete independently. 

If we talking about the driver/executor pods, yes, they complete independently and don't require Spark operator instance to perform status updates. However, `WaitForJobState` method is checking the status of `SparkApplication` resource, not a `Pod`, and the former is updated by the operator controller of a new leader. 
But I agree, that checking the logs to verify the newly elected leader handles job status update would be more explicit for this case.

akirillov: SGTM, thanks for the clarification 👍

alembiewski:
Thanks for the review, @akirillov! Going to merge this PR after https://github.com/kudobuilder/operators/pull/167 is closed.

PR #76, opened 2019-12-20
[DCOS-61939] Add integration test for Volcano scheduler

alembiewski:
Resolves [DCOS-61939: Integration test and documentation for Volcano integration](https://jira.mesosphere.com/browse/DCOS-61939)
Adds a test for exercising integration with Volcano (https://github.com/volcano-sh/volcano)

## Why are the changes needed?
To verify `SparkApplication` runs and completes successfully when using Volcano as a batch scheduler.

test from this repo

### DIFF
--- a/tests/volcano_test.go
+++ b/tests/volcano_test.go
@@ -1,0 +1,84 @@
+package tests
+
+import (
+	"fmt"
+	"github.com/mesosphere/kudo-spark-operator/tests/utils"
+	"github.com/stretchr/testify/suite"
+	"testing"
+)
+
+const volcanoInstallerUrl = "https://raw.githubusercontent.com/volcano-sh/volcano/release-0.2/installer/volcano-development.yaml"
+
+type VolcanoIntegrationTestSuite struct {
+	operator utils.SparkOperatorInstallation
+	suite.Suite
+}
+
+func TestVolcanoSuite(t *testing.T) {
+	suite.Run(t, new(VolcanoIntegrationTestSuite))
+}
+
+func (suite *VolcanoIntegrationTestSuite) SetupSuite() {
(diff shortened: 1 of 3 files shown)


tests/volcano_test.go
```diff
@@ -0,0 +1,84 @@
+package tests
+
+import (
+	"fmt"
+	"github.com/mesosphere/kudo-spark-operator/tests/utils"
+	"github.com/stretchr/testify/suite"
+	"testing"
+)
+
+const volcanoInstallerUrl = "https://raw.githubusercontent.com/volcano-sh/volcano/release-0.2/installer/volcano-development.yaml"
```
akirillov: is it worth maintaining a copy of this spec in test resource folder instead of downloading it?

tests/volcano_test.go
```diff
@@ -0,0 +1,84 @@
...
+	if err := suite.operator.WaitUntilSucceeded(sparkPi); err != nil {
+		suite.FailNow(err.Error())
+	}
+
+	// assert that the driver pod was scheduled by volcano
+	driverPodName := utils.DriverPodName(jobName)
+	component, err := utils.Kubectl("get", "events",
+		"--namespace", sparkPi.Namespace,
+		"--field-selector", fmt.Sprint("involvedObject.name=", driverPodName),
+		"-o", "jsonpath={.items[0].source.component}")
```
akirillov: this probably deserves an extended comment for developers unfamiliar with the topic. Would be great to add a line about why this check is sufficient and why this field is different when using Volcano.

tests/volcano_test.go
alembiewski: yes, makes sense, switched to using local copy

PR #80, opened 2020-01-04
Fix for S3 folder removal in Spark History test

alembiewski:
This PR aims to fix SHS test failures caused by S3 folder deletion when parallel builds are running in CI

Currently, an S3 folder, used as a log directory for storing Spark applications, is removed when the test starts and finishes. This approach must not be used in case when multiple builds are running in parallel in CI. When two tests are running concurrently, one could cause another one to fail due to folder clean-up (example [build log](https://teamcity.mesosphere.io/viewLog.html?tab=buildLog&logTab=tree&filter=debug&expand=all&buildId=2525628&_focus=5696))

The proposed solution is to delegate the clean-up process to the S3 lifecycle management feature and remove clean-up from the test code.

Expiration for `kudo-ds-ci-artifacts/kudo-spark-operator/tests/spark-history-server/*` is set to one day.

test from this repo

### DIFF
--- a/tests/basic_test.go
+++ b/tests/basic_test.go
@@ -2,14 +2,12 @@ package tests
 
 import (
 	"fmt"
-	"os"
-	"strings"
-	"testing"
-	"time"
-
 	"github.com/mesosphere/kudo-spark-operator/tests/utils"
 	log "github.com/sirupsen/logrus"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"os"
+	"strings"
+	"testing"
 )
 
 func TestMain(m *testing.M) {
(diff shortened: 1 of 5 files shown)


tests/basic_test.go
```diff
@@ -96,13 +94,8 @@ func TestSparkHistoryServerInstallation(t *testing.T) {
 		t.Fatal("AWS_BUCKET_PATH is not configured")
 	}
 
-	// Make sure folder is deleted
-	err := utils.AwsS3DeleteFolder(awsBucketName, awsFolderPath)
```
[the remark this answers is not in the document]
alembiewski: Yes, it definitely reasonable to do so. Switched to the proposed solution.

tests/basic_test.go
```diff
@@ -207,6 +202,7 @@ func TestSparkHistoryServerInstallation(t *testing.T) {
 		log.Infof("Spark History Server logs:")
 		utils.Kubectl("logs", "-n", spark.Namespace, historyServerPodName)
 	}
+
 	utils.AwsS3DeleteFolder(awsBucketName, awsFolderPath)
```
akirillov: will this function be called disregarding the previous errors? can we use it in `defer` after the folder is created?

alembiewski: it would make sense to keep this files in case of errors for troubleshooting, but we already collecting logs if something goes wrong, so yes, let's change that

PR #81, opened 2020-01-09
Stabilize HA tests suite

alembiewski:
This PR aims to fix flakiness in HA Test suite.

`TestHASuite/TestFailover` test sometimes fails because of a missing leader pod [(example](https://teamcity.mesosphere.io/viewLog.html?buildId=2532422&tab=buildResultsDiv&buildTypeId=Frameworks_DataServices_Kudo_Spark_Pr_SparkPrKonvoyKudoV080pre2))
As it's turned out, at some point during deployment of the operator, the instance of `ReplicaSet` gets recreated (probably, the initial one has failed), causing previously deployed pods to be deleted. The following changes were introduced to stabilize to test flow:

- `kubectl wait` command was added to `SetupSuite` to explicitly check whether the deployment is in the completed state;
-  `getLeaderElectionRecord` method was updated to check the leader pod exists;
- logically reordered tests in the suite - it makes sense testing configuration and leader election before the failover scenario;
- added deployment state to test log for troubleshooting.

- locally on private cluster;
- tests from this repo;

### DIFF
--- a/tests/ha_test.go
+++ b/tests/ha_test.go
@@ -21,6 +21,7 @@ import (
 const electionRecordRetryInterval = 2 * time.Second
 const electionRecordRetryTimeout = 30 * time.Second
 const processingKeyLogRecordFormat = "Starting processing key: \"%s/%s\""
+const deploymentWaitTimeout = 1 * time.Minute
 
 type LeaderElectionParameters struct {
 	Replicas                    int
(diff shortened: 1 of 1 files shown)


tests/ha_test.go
```diff
@@ -61,14 +61,15 @@ func (suite *HighAvailabilityTestSuite) SetupSuite() {
 	if err := suite.operator.InstallSparkOperator(); err != nil {
 		suite.FailNow(err.Error())
 	}
-
+	utils.Kubectl("wait", "deployment", "--all", "--for", "condition=available",
+		"--namespace", suite.operator.Namespace, "--timeout=60s")
```
akirillov: nit: should the timeout be configurable?

alembiewski:
Thanks for the review @akirillov! Made a timeout configurable as per your suggestion. CI check for kudo 0.8.0 is successful, merging.

PR #82, opened 2020-01-13
Bump Kudo version to 0.10.0

alembiewski:
Switch to the latest Kudo v0.10.0 

This update will make it possible to install monitoring resources automatically with Spark Operator, introduces new plans support, as well as cli improvements and bugfixes.

tests from this repo

### DIFF
--- a/tests/tenancy_test.go
+++ b/tests/tenancy_test.go
@@ -11,7 +11,6 @@ import (
 )
 
 func TestTenancyTwoOperatorsDifferentNamespaces(t *testing.T) {
-	t.Skip("Test is temporarily disabled due to KUDO v0.10.0 incompatibility")
 	operators := operatorBuilder(2, true, true)
 	for _, operator := range operators {
 		err := operator.InstallSparkOperator()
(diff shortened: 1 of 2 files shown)


alembiewski:
Blocked until this issue is resolved: https://github.com/kudobuilder/kudo/issues/1265

alembiewski:
Based on CI check, it looks like a fix made in https://github.com/kudobuilder/operators/pull/195 partially resolves https://github.com/kudobuilder/kudo/issues/1265, but only for single-instance deployments. In multi-tenant mode (two or more operators are deployed), the same error is still reported.

alembiewski:
CI check passed after disabling multi-tenancy tests. Will enable them back when an issue with installing multiple operators is resolved.

tests/tenancy_test.go
```diff
@@ -11,6 +11,7 @@ import (
 )
 
 func TestTenancyTwoOperatorsDifferentNamespaces(t *testing.T) {
+	t.Skip("Test is temporarily disabled due to KUDO v0.10.0 incompatibility")
```
akirillov: what if instead of skipping the whole test, we install the second instance with `installCrds=false` and `createRBAC=false` so that we don't install cluster-scoped resources?

alembiewski: I tried this approach and it worked even without disabling RBAC (which is strange, since `ClusterRole` and `ClusterRoleBinding` are also cluster-scoped resources), but tests are still failing: `TestCRDsDeletion` fails because now CRDs are deleted together with the first operator, even though the second operator is still running.

PR #83, opened 2020-01-14
Scale test scripts updates and improvements

akirillov:
This PR contains updated automation scripts and configuration for the scale test workloads.

* to exclude monitoring resources which are part of the operator installation
* add wait timeout on the installation to allow finer-grained troubleshooting procedures
* to tweak TeraSort job resource/scale parameters so it is able to complete successfully
* update Grafana dashboard to return missing metrics 

* the changes were tested during periodic scale testing at D2iQ


akirillov:
good point, @alembiewski. I think by the time of the next MWT we'll have some other upgrades so let's address them closer to that date.

PR #84, opened 2020-01-24
Test metrics in Prometheus

alembiewski:
Resolves [DCOS-62599](https://jira.mesosphere.com/browse/DCOS-62599).
This PR introduces a test that verifies all the metrics used in the dashboards are scraped and reported to Prometheus.

The test automatically checks the following components are working properly:
- monitoring resources are deployed;
- metrics are reported by Operator and Spark Applications and scraped by Prometheus;
- Grafana dashboards, by executing real-world queries in Prometheus;

tests from this repo


tests/metrics_test.go
```diff
@@ -0,0 +1,263 @@
...
+	"strings"
+	"testing"
+	"time"
+)
+
+const dashboardsDir = "../operators/repository/spark/docs/latest/resources/dashboards"
+const jobName = "mock-task-runner"
+const jobTemplate = "spark-mock-task-runner-with-monitoring.yaml"
+const prometheusNamespace = "kubeaddons"
+const prometheusPod = "prometheus-prometheus-kubeaddons-prom-prometheus-0"
```
akirillov: Is there a safer way to discover Prometheus pod name instead of hardcoding the pod name? It is unlikely but quite possible that it can change after Prometheus Operator upgrade.

alembiewski:
Thank you for the review, @akirillov! 
Regarding failed application - yes, it is intentional, I emulate failure by misconfiguring application arguments:
https://github.com/mesosphere/kudo-spark-operator/blob/ac757bad17325f756b196b647f256e30f0f80e44/tests/metrics_test.go#L181
It's necessary to submit two applications for the test (successful and failed). Otherwise, some metrics will not be reported and we will not be able to verify all the queries.

tests/metrics_test.go
alembiewski: good point, will try to obtain pod name at runtime via label selector

PR #88, opened 2020-02-03
D2IQ-63474: Integration test and docs for S3 access credentials

alembiewski:
Resolves [D2IQ-63474: Integration test and docs for S3 access credentials](https://jira.d2iq.com/browse/D2IQ-63474).
This PR adds an integration test for verifying access to S3 from Spark job and Spark History Server using AWS credentials obtained from `Secrets`. Also, it modifies the existing test for Spark History server to adhere to the new method of handling credentials.

- To check AWS credentials are propagated to Spark applications from a Secret
- To verify a new method of mounting Spark config file containing AWS credentials to the Spark History Server as a Secret, which name is passed via `historyServerSparkConfSecret` KUDO parameter.

PR for docs and updated configuration: https://github.com/kudobuilder/operators/pull/214 
Test job can be found here: https://github.com/mesosphere/spark-build/blob/master/tests/jobs/scala/src/main/scala/S3Job.scala

tests from this repo.

### DIFF
--- a/tests/templates/spark-linear-regression-history-server-job.yaml
+++ b/tests/templates/spark-linear-regression-history-server-job.yaml
@@ -61,3 +61,20 @@ spec:
     memory: "512m"
     labels:
       version: {{ .SparkVersion }}
+    env:
+      - name: AWS_ACCESS_KEY_ID
+        valueFrom:
+          secretKeyRef:
+            name: {{ .Params.AwsSecretName }}
+            key: AWS_ACCESS_KEY_ID
+      - name: AWS_SECRET_ACCESS_KEY
+        valueFrom:
+          secretKeyRef:
+            name: {{ .Params.AwsSecretName }}
+            key: AWS_SECRET_ACCESS_KEY
+      - name: AWS_SESSION_TOKEN
+        valueFrom:
+          secretKeyRef:
+            name: {{ .Params.AwsSecretName }}
+            key: AWS_SESSION_TOKEN
+            optional: true
--- a/tests/templates/spark-linear-regression-history-server-job.yaml
+++ b/tests/templates/spark-linear-regression-history-server-job.yaml
@@ -61,20 +61,3 @@ spec:
     memory: "512m"
     labels:
       version: {{ .SparkVersion }}
-    env:
-      - name: AWS_ACCESS_KEY_ID
-        valueFrom:
-          secretKeyRef:
-            name: {{ .Params.AwsSecretName }}
-            key: AWS_ACCESS_KEY_ID
-      - name: AWS_SECRET_ACCESS_KEY
-        valueFrom:
-          secretKeyRef:
-            name: {{ .Params.AwsSecretName }}
-            key: AWS_SECRET_ACCESS_KEY
-      - name: AWS_SESSION_TOKEN
-        valueFrom:
-          secretKeyRef:
-            name: {{ .Params.AwsSecretName }}
-            key: AWS_SESSION_TOKEN
-            optional: true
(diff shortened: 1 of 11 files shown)


alembiewski:
Yes, I agree. Having a more generic way to specify secrets and configuration parameters is more beneficial in this case. I introduced the new parameter, `historyServerSparkConfSecret`, which allows users to pass the name of a secret with `spark-default.conf`, that will be mounted to default conf location in SHS pod. WDYT?

tests/templates/spark-linear-regression-history-server-job.yaml
```diff
@@ -44,6 +38,23 @@ spec:
...
+            key: AWS_ACCESS_KEY_ID
+      - name: AWS_SECRET_ACCESS_KEY
+        valueFrom:
+          secretKeyRef:
+            name: {{ .Params.AwsSecretName }}
+            key: AWS_SECRET_ACCESS_KEY
+      - name: AWS_SESSION_TOKEN
+        valueFrom:
+          secretKeyRef:
+            name: {{ .Params.AwsSecretName }}
```
akirillov: I think `.Params.AwsSecretName` contradicts the approach described in https://github.com/kudobuilder/operators/pull/214, doesn't it?

tests/templates/spark-s3-readwrite.yaml
```diff
@@ -0,0 +1,56 @@
...
+            key: AWS_ACCESS_KEY_ID
+      - name: AWS_SECRET_ACCESS_KEY
+        valueFrom:
+          secretKeyRef:
+            name: {{ .Params.AwsSecretName }}
+            key: AWS_SECRET_ACCESS_KEY
+      - name: AWS_SESSION_TOKEN
+        valueFrom:
+          secretKeyRef:
+            name: {{ .Params.AwsSecretName }}
```
akirillov: ditto

tests/templates/spark-linear-regression-history-server-job.yaml
alembiewski: No, it is actually two different approaches: the one described here https://github.com/kudobuilder/operators/pull/214 is based on mounting a `spark-defaults.conf`, which contains `spark.hadoop.fs.s3a.*` properties, that will be picked up by SHS.
This one reflects the way we currently inject secrets into `SparkApplication` because of [this issue](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/216) - as for now,  it is possible to provide Spark properties only via `spec.sparkConf`, but it will cause exposure of sensitive data in the application spec, so we are using a separate secret to provide keys and passwords to Spark app, e.g:
```yaml
apiVersion: v1
data:
  AWS_ACCESS_KEY_ID: <base64-encoded>
  AWS_SECRET_ACCESS_KEY: <base64-encoded>
kind: Secret
metadata:
  name: spark-secrets
type: Opaque
```

akirillov: Thanks for the clarification. My point here was more about why do we need an Operator parameter if we have to specify it on per-job basis? As this parameter is not applied automatically for all submitted jobs - there doesn't seem to be much sense of including it to into Operator configuration.

alembiewski: They are two different parameters: `historyServerSparkConfSecret ` used in SHS only, and this `.Params.AwsSecretName ` is name of a secret used for Spark job in the test. Ideally, we should have I single Secret with spark properties, which would be shared between all the components (SHS, SparkApp), but as I said, currently there is no way to pass spark configuration to `SparkApplication` via conf file, so that’s why we create a separate secret with `AWS_*` credentials and use in Spark jobs

akirillov: Got it, thanks! So it is a parameter for tests only which makes sense.

PR #90, opened 2020-02-07
D2IQ-63472: Add an integration test for RPC network encryption and TLS

alembiewski:
Resolves [D2IQ-63472: Add an integration test for RPC network encryption and SSL](https://jira.d2iq.com/browse/D2IQ-63472)

This PR introduces changes to `entrypoint.sh` to add the possibility to configure Spark security properties via K8s Secrets
Also, it adds an integration test to verify introduced changes by running test applications with different config sets (RPC auth + network encryption, SSL using a self-signed certificate).

To verify Spark Security features work for Kudo Spark Operator.

tests from this repo.

**Note:** this PR relies on a changes, introduced in https://github.com/mesosphere/kudo-spark-operator/pull/88, therefore, in order to pass CI check, it needs to be rebased ofter aforementioned PR is merged.


akirillov:
Took a quick look at this PR - I think the approach is aligned with what has been proposed in https://github.com/mesosphere/kudo-spark-operator/pull/88 and looks okay for Spark 2.4.x series.

Makefile
```diff
@@ -7,7 +7,7 @@ SCRIPTS_DIR := $(ROOT_DIR)/scripts
 KUDO_TOOLS_DIR := $(ROOT_DIR)/shared
 SPARK_OPERATOR_DIR := $(ROOT_DIR)/spark-on-k8s-operator
 
-export KONVOY_VERSION ?= v1.1.5
+export KONVOY_VERSION ?= v1.2.4
```
akirillov: out of curiosity, why not 1.3.0?

tests/encryption_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+	// when ssl is configured, check Spark UI is accessible via https on 4440 port
+	if _, present := sparkConf["spark.ssl.enabled"]; present {
+		checkSparkUI(appName, sparkApp, suite)
+	}
+
+	suite.Assert().NoError(suite.operator.WaitUntilSucceeded(sparkApp))
+}
+
+func checkSparkUI(appName string, sparkApp utils.SparkJob, suite *SparkEncryptionSuite) {
```
akirillov: 👍

tests/encryption_test.go
```diff
@@ -0,0 +1,239 @@
...
+// returns modified Spark configuration containing 'JAVA_TOOL_OPTIONS' set for executor env,
+// which allows to pass parameters to executor processes in order to setup RPC auth and encryption.
+// under the normal circumstances, all the properties should be copied from a driver configuration,
+// but for some reason it is not working when running on k8s, so this method can be considered as a temporary workaround.
+func copyConfToExecutorEnv(sparkConf map[string]string) map[string]string {
+	var javaOpts []string
+	for key, value := range sparkConf {
+		javaOpts = append(javaOpts, fmt.Sprintf("-D%s=%s", key, value))
+	}
+	sparkConf["spark.executorEnv.JAVA_TOOL_OPTIONS"] = strings.Join(javaOpts, " ")
```
akirillov: IIUC, this step must be performed by users in order to provide secrets to executors. Can we instead handle this situation in `entrypoint.sh` and set `--conf spark.kubernetes.executor.secretKeyRef.ENV_NAME=name:key` when driver is launched? Accordingly, we can add environment variables handling in `entripoint.sh` of the executor in a similar way we do for the Driver.

WDYT?

alembiewski: This is already handled here:
https://github.com/mesosphere/kudo-spark-operator/blob/7b9d274508e93e8d78d467b10f1dc698a33d2e12/tests/templates/spark-mock-task-runner-encrypted.yaml#L63
`JAVA_TOOL_OPTIONS` here is used to pass Spark configuration parameters (`spark.authenticate`, `spark.network.crypto.enabled` etc) to the executors,  which I believe should be [copied from the driver](https://github.com/apache/spark/blob/v2.4.4/core/src/main/scala/org/apache/spark/executor/CoarseGrainedExecutorBackend.scala#L188-L215), but for some reason, they don't.

Makefile
alembiewski: I used the default version, specified here: https://github.com/mesosphere/data-services-kudo/blob/ae0d6d4e4af5e24575841fcf9615ab4f14d93f4c/vendor/konvoy.sh#L5
But I think we can bump it to `1.3.0`, let's see how it goes

tests/encryption_test.go
alembiewski: Ok, I think I got it: executor fails to fetch configuration because it can't authenticate with a driver itself. Logs:
```
20/02/13 09:16:38 DEBUG UserGroupInformation: PrivilegedActionException as:root (auth:SIMPLE) cause:org.apache.spark.SparkException: Exception thrown in awaitResult: 
Exception in thread "main" java.lang.reflect.UndeclaredThrowableException
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1911)
	at org.apache.spark.deploy.SparkHadoopUtil.runAsSparkUser(SparkHadoopUtil.scala:64)
	at org.apache.spark.executor.CoarseGrainedExecutorBackend$.run(CoarseGrainedExecutorBackend.scala:188)
	at org.apache.spark.executor.CoarseGrainedExecutorBackend$.main(CoarseGrainedExecutorBackend.scala:281)
	at org.apache.spark.executor.CoarseGrainedExecutorBackend.main(CoarseGrainedExecutorBackend.scala)
Caused by: org.apache.spark.SparkException: Exception thrown in awaitResult: 
	at org.apache.spark.util.ThreadUtils$.awaitResult(ThreadUtils.scala:226)
	at org.apache.spark.rpc.RpcTimeout.awaitResult(RpcTimeout.scala:75)
	at org.apache.spark.rpc.RpcEnv.setupEndpointRefByURI(RpcEnv.scala:101)
	at org.apache.spark.executor.CoarseGrainedExecutorBackend$$anonfun$run$1.apply$mcV$sp(CoarseGrainedExecutorBackend.scala:201)
	at org.apache.spark.deploy.SparkHadoopUtil$$anon$2.run(SparkHadoopUtil.scala:65)
	at org.apache.spark.deploy.SparkHadoopUtil$$anon$2.run(SparkHadoopUtil.scala:64)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:422)
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1893)
	... 4 more
Caused by: java.lang.RuntimeException: java.lang.IllegalStateException: Expected SaslMessage, received something else (maybe your client does not have SASL enabled?)
```



So it is sufficient to only pass `spark.authenticate=true` together with `_SPARK_AUTH_SECRET` environment variable to the executor. When those settings are set, the executor will be able to authenticate with the driver and fetch the rest of the configuration. 
So basically, we can replace `JAVA_TOOL_OPTIONS` with a single env variable in the executor spec, e.g;
```yaml
spec:
  executor:
    env:
      - name: SPARK_AUTHENTICATE
        value: true
...
We also need to modify `entrypoint.sh` to pass the required property to the executor via `SPARK_EXECUTOR_JAVA_OPTS` based on this variable.

Comment (alembiewski) [reply tests/encryption_test.go]: OR we can just check if `_SPARK_AUTH_SECRET` is set and enable `spark.authenticate` on executors based on that variable. I would go with this approach since it will require less configuration in the application spec. WDYT?

**Comment by alembiewski** [comment]
@akirillov, I excepted the usage of `JAVA_TOOL_OPTIONS` for passing RPC configuration parameters to executors as proposed here: https://github.com/mesosphere/kudo-spark-operator/pull/90#discussion_r378782705
CI check will not pass until #88 is merged to master. It should be fine after the rebase.

File: tests/templates/spark-mock-task-runner-encrypted.yaml [hunk 4]
```diff
@@ -0,0 +1,76 @@
...
+    labels:
+      version: {{ .SparkVersion }}
+    serviceAccount: {{ .ServiceAccount }}
+    {{- if .Params.SslEnabled }}
+    secrets:
+      - name: {{ .Params.SslSecrets }}
+        path: "/tmp/spark/ssl"
+        secretType: Generic
+    {{- end }}
+    env:
```
akirillov: can we check, if it is possible to specify the environment in a less verbose way via `sparkConf`? E.g.:
```
sparkConf:
  "spark.kubernetes.driver.secretKeyRef.SPARK_AUTHENTICATE_SECRET": "{{ .Params.SparkSecrets }}:auth-secret"
  "spark.kubernetes.driver.secretKeyRef.SPARK_SSL_KEYPASSWORD": "{{ .Params.SparkSecrets }}:key-password"
  "spark.kubernetes.driver.secretKeyRef.SPARK_SSL_KEYSTOREPASSWORD": "{{ .Params.SparkSecrets }}:keystore-password"
  "spark.kubernetes.driver.secretKeyRef.SPARK_SSL_TRUSTSTOREPASSWORD": "{{ .Params.SparkSecrets }}:truststore-password"
  "spark.kubernetes.executor.secretKeyRef.SPARK_AUTHENTICATE_SECRET": "{{ .Params.SparkSecrets }}:auth-secret"
```

images/spark/conf/entrypoint.sh
```diff
@@ -0,0 +1,157 @@
...
+    add_spark_conf_if_non_empty "spark.authenticate.secret" "${SPARK_AUTHENTICATE_SECRET}"
+    add_spark_conf_if_non_empty "spark.ssl.keyPassword" "${SPARK_SSL_KEYPASSWORD}"
+    add_spark_conf_if_non_empty "spark.ssl.keyStorePassword" "${SPARK_SSL_KEYSTOREPASSWORD}"
+    add_spark_conf_if_non_empty "spark.ssl.trustStorePassword" "${SPARK_SSL_TRUSTSTOREPASSWORD}"
+}
+
+configure_spark_properties
+
+# if _SPARK_AUTH_SECRET is set, enable RPC authetication for executors
+if [[ -n "${_SPARK_AUTH_SECRET}" ]]; then
```
akirillov: I'd suggest using the same variable as for the driver: `SPARK_AUTHENTICATE_SECRET` and set `_SPARK_AUTH_SECRET` in this script. That way it will be easier to describe it in the documentation, e.g.:
```
In order to enable RPC authentication:
* set `spark.authenticate` configuration property to `true` 
* mount `SPARK_AUTHENTICATE_SECRET` environment variable from a secret for both the Driver and Executors

Here's an example Spark Application YAML:
<here goes the example>
```

alembiewski:
@akirillov thank you for the review! I added a few more commits addressing the comments, please take a look. I will also update the docs to reflect these changes.

images/spark/conf/entrypoint.sh
```diff
@@ -0,0 +1,159 @@
...
+    add_spark_conf_if_non_empty "spark.ssl.keyStorePassword" "${SPARK_SSL_KEYSTOREPASSWORD}"
+    add_spark_conf_if_non_empty "spark.ssl.trustStorePassword" "${SPARK_SSL_TRUSTSTOREPASSWORD}"
+}
+
+configure_spark_properties
+
+# if SPARK_AUTHENTICATE_SECRET is set, enable RPC authetication for executors
+# and provide the auth secret via _SPARK_AUTH_SECRET variable
+if [[ -n "${SPARK_AUTHENTICATE_SECRET}" ]]; then
+  SPARK_EXECUTOR_JAVA_OPTS+=("-Dspark.authenticate=true")
```
akirillov: TBH, I'm not sure we should automate this for the users (because essentially we're hiding it by making it implicit) but not a blocker anyway

tests/encryption_test.go
```diff
@@ -0,0 +1,240 @@
...
+
+func (suite *SparkEncryptionSuite) TestSSL() {
+	sparkConf := map[string]string{
+		"spark.ssl.enabled":    "true",
+		"spark.ssl.keyStore":   suite.keyStorePath,
+		"spark.ssl.protocol":   "TLSv1.2",
+		"spark.ssl.trustStore": suite.trustStorePath,
+		"spark.kubernetes.driver.secretKeyRef.SPARK_SSL_KEYPASSWORD":        fmt.Sprintf("%s:key-password", suite.sparkSecrets),
+		"spark.kubernetes.driver.secretKeyRef.SPARK_SSL_KEYSTOREPASSWORD":   fmt.Sprintf("%s:keystore-password", suite.sparkSecrets),
+		"spark.kubernetes.driver.secretKeyRef.SPARK_SSL_TRUSTSTOREPASSWORD": fmt.Sprintf("%s:truststore-password", suite.sparkSecrets),
```
akirillov: 👍

tests/encryption_test.go
```diff
@@ -0,0 +1,240 @@
...
+
+func (suite *SparkEncryptionSuite) TearDownSuite() {
+	suite.operator.CleanUp()
+}
+
+func (suite *SparkEncryptionSuite) TestRpc() {
+	sparkConf := map[string]string{
+		"spark.authenticate": "true",
+		"spark.kubernetes.driver.secretKeyRef.SPARK_AUTHENTICATE_SECRET":   fmt.Sprintf("%s:auth-secret", suite.sparkSecrets),
+		"spark.kubernetes.executor.secretKeyRef.SPARK_AUTHENTICATE_SECRET": fmt.Sprintf("%s:auth-secret", suite.sparkSecrets),
```
akirillov: 👍

images/spark/conf/entrypoint.sh
alembiewski: I agree, but I think this is more like an implementation detail. If I, as a user, specify `spark.authenticate: true` in `sparkConf` for my application, I would expect the authentication will be enabled for driver and executors automatically, without any explicit configuration for the executor, except for the auth secret (and it is aligned with [the documentation](https://spark.apache.org/docs/latest/security.html#authentication)).

PR #92, opened 2020-02-19
D2IQ-62600: Integration test for Spark with HDFS and Kerberos

alembiewski:
Resolves [D2IQ-62600: Support Kerberos security](https://jira.d2iq.com/browse/D2IQ-62600).
This PR adds a test suite to check the integration of Spark Operator, including Spark History Server, with HDFS, secured via Kerberos. 

To automatically check Spark applications can access HDFS in a secured environment, as well as Spark History Server can use HDFS as event log storage with Spark Operator. 

tests from this repo

### DIFF
--- a/tests/hdfs_kerberos_test.go
+++ b/tests/hdfs_kerberos_test.go
@@ -62,7 +62,6 @@ func TestHdfsIntegrationSuite(t *testing.T) {
 }
 
 func (suite *HdfsIntegrationSuite) SetupSuite() {
-	utils.Kubectl("delete", "ns", namespace)
 	if _, err := utils.Kubectl("create", "ns", namespace); err != nil {
 		suite.FailNow("Error while creating namespace", err)
 	}
--- a/tests/hdfs_kerberos_test.go
+++ b/tests/hdfs_kerberos_test.go
@@ -1,0 +1,150 @@
+package tests
+
+import (
+	"fmt"
+	"github.com/mesosphere/kudo-spark-operator/tests/utils"
+	"github.com/stretchr/testify/suite"
+	"testing"
+	"time"
+)
+
+var (
+	resourceFolder         = "resources/hdfs-kerberos"
+	namespace              = "hdfs-kerberos"
+	hdfsKerberosDeployment = []Resource{
+		{
+			name: "configmaps",
+			path: "configmaps",
+		},
+		{
+			name: "volumes",
+			path: "volumes",
+		},
+		{
+			name: "kerberos",
+			path: "kerberos-deployment.yaml",
+			wait: true,
+		},
+		{
+			name: "hdfs-namenode",
+			path: "namenode-deployment.yaml",
+			wait: true,
+		},
+		{
+			name: "hdfs-datanode",
+			path: "datanode-deployment.yaml",
+			wait: true,
+		},
+	}
+	eventLogDir       = "hdfs://namenode.hdfs-kerberos.svc.cluster.local:9000/history"
+	hadoopTokenSecret = "hadoop-token"
+)
+
+type Resource struct {
+	name string
+	path string
+	wait bool
+}
+
+type HdfsIntegrationSuite struct {
+	operator utils.SparkOperatorInstallation
+	suite.Suite
+}
+
+func TestHdfsIntegrationSuite(t *testing.T) {
+	suite.Run(t, new(HdfsIntegrationSuite))
+}
+
+func (suite *HdfsIntegrationSuite) SetupSuite() {
+	utils.Kubectl("create", "ns", namespace)
+	// deploy KDC and HDFS
+	for _, resource := range hdfsKerberosDeployment {
+		utils.Kubectl("apply", "-f", fmt.Sprint(resourceFolder, "/", resource.path), "-n", namespace)
+
+		if resource.wait {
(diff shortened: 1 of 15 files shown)


tests/hdfs_kerberos_test.go
```diff
@@ -0,0 +1,150 @@
...
+		"--output=jsonpath={.items[*].metadata.name}", "-n", namespace)
+	if err != nil {
+		suite.FailNow("Error while getting a name of Namenode pod", err)
+	}
+	// run init script to copy test data to HDFS and export delegation token
+	utils.Kubectl("exec", nameNodePod, "-n", namespace, "--", "init.sh")
+
+	// copy delegation token from the pod to a local filesystem
+	utils.Kubectl("cp", fmt.Sprint(nameNodePod, ":", "tmp/hadoop.token"), "/tmp/hadoop.token", "-n", namespace)
+	utils.Kubectl("create", "secret", "generic", hadoopTokenSecret, "--from-file", "/tmp/hadoop.token", "-n", namespace)
```
akirillov: Looking through the changes in SHS deployment template and it looks like that Operator assumes that the file name in the secret is`hadoop.token` - we should reflect it in the docs.

tests/hdfs_kerberos_test.go
```diff
@@ -0,0 +1,150 @@
...
+		suite.FailNow(err.Error())
+	}
+}
+
+func (suite *HdfsIntegrationSuite) AfterTest(suiteName, testName string) {
+	utils.Kubectl("kudo", "uninstall",
+		"--instance", suite.operator.InstanceName,
+		"--namespace", suite.operator.Namespace)
+	// TODO: temporary workaround: explicitly waiting for uninstall to clean all the resources before the next test starts.
+	time.Sleep(15 * time.Second)
```
akirillov: can this workaround be replaced with periodic polling with timeout?

tests/hdfs_kerberos_test.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func (suite *HdfsIntegrationSuite) Test_Spark_Hdfs_Kerberos() {
+	jobName := "spark-hdfs-kerberos"
+	sparkPi := utils.SparkJob{
```
akirillov: aren't we using `HdfsTest` job instead of Pi? Let's update the variable name to avoid the confusion

tests/resources/hdfs-kerberos/datanode-deployment.yaml
```diff
@@ -0,0 +1,57 @@
...
+        pod.beta.kubernetes.io/hostname: datanode
+      labels:
+        name: hdfs-datanode
+        kerberosService: datanode
+    spec:
+      containers:
+      - command:
+        - start-datanode.sh
+        name: datanode
+        image: lembiewski/hdfs-kerberos:v1.0.0
```
akirillov: I think it makes sense to provide source Dockerfiles so that we can easily rebuild the images in case the repo changes. Although we should find a proper repo for the images.

alembiewski: done

tests/hdfs_kerberos_test.go
alembiewski: fixed

tests/hdfs_kerberos_test.go
alembiewski: The issue with `kudo uninstall` is that we can't actually verify all the related resources are deleted. What I observed locally, is sometimes the second test failed because of the incomplete cleanup after `uninstall`, meaning when the next installation starts, some resources of the previous one are still not cleaned up, whick cause deploy plan failure.
So, as it was suggested in [Slack thread](https://mesosphere.slack.com/archives/CR1SVH3PD/p1582186988020600), one of the possible solution is to delete the instance itself and together with `--wait` flag.

tests/hdfs_kerberos_test.go
alembiewski: added explicit comment to the test code about `hadoop.token` secret key

tests/hdfs_kerberos_test.go
alembiewski: Looking at the failed build history, I would say a solution with the `wait` flag could not be considered reliable. I going to switch back to the default `operator.Cleanup()` approach, but I need to make some changes and separate the Operator and HDFS with Kerberos using different namespaces. This will allow us to clean operators and preserve HDFS installation.

tests/resources/hdfs-kerberos/namenode-deployment.yaml
```diff
@@ -0,0 +1,79 @@
...
+        configMap:
+          name: hadoop-conf
+      - name: krb-config
+        configMap:
+          name: krb5
+      # NOTE: by default, Konvoy on AWS uses 'aws-ebs-csi-driver', which is a CSI driver for Amazon EBS,
+      # and, according to https://kubernetes.io/docs/concepts/storage/persistent-volumes/, AWSElasticBlockStore doesn't support
+      # ReadWriteMany mode for PVs. That means a volume can't be mounted as read-write by many nodes.
+      # The solution is to use pod affinity in order to place all Kerberos and HDFS pods on the same node
+      # and use ReadWriteOnce mode fo PV.
```
akirillov: 👍

PR #117, opened 2021-01-28
Update scale test for MWT-10 Dry Run

alexeygorobets:
To use the latest Kudo Spark in the MWT

## Tolerations
To tolerate taints use the following toleration for driver and executor:
```yaml
tolerations:
   - key: "dedicated"
     operator: "Equal"
     value: "Cassandra"
     effect: "NoExecute"```

### CODE REVIEW

File: scale-tests/scripts/install.sh [hunk 1]
```diff
@@ -26,7 +26,7 @@ for i in $(seq ${1}); do
     sed 's|SPARK_NAMESPACE|'"${NAMESPACE}"'|g' ${TEMPLATES_DIR}/namespace.tmpl | kubectl apply -f -
     sed 's|SERVICE_ACCOUNT_NAME|'"${SERVICE_ACCOUNT_NAME}"'|g' ${TEMPLATES_DIR}/service-account.tmpl | kubectl apply --namespace "${NAMESPACE}" -f -
 
-    kubectl kudo --namespace "${NAMESPACE}" install --instance "${INSTANCE_NAME_PREFIX}-${i}" "${OPERATOR_DIR}" \
+    kubectl kudo --namespace "${NAMESPACE}" install --instance "${INSTANCE_NAME_PREFIX}-${i}" spark \
```
akirillov, line 29: is it intentional?

akirillov (approved):
thanks, @alexeygorobets. Have one nitpick, otherwise LGTM

scale-tests/scripts/install.sh
alexeygorobets: I've done it to install spark from kudo registry, will return it back

